PR #49, opened 2018-05-24
Fancy ticks ••⦿••

eddanger:
The ticks are a lot more fancy.

screen shot 2018-05-23 at 10 21 23 pm


senorprogrammer:
I like this idea so much I incorporated it in as a configurable setting. See this commit: https://github.com/senorprogrammer/wtf/commit/5f7c3260c1ac57db4b4a6e00ead627dd96a516f5

eddanger:
Sounds good! 👍 My code will live on... in the cloud ☁️

Issue #11, opened 2018-04-15
Watch config file for changes


senorprogrammer:
Refresh app and widgets when config file changes.

PR #61, opened 2018-05-28
Watch config file for changes

jeangovil:
Issue: #11

### DIFF
--- a/wtf.go
+++ b/wtf.go
@@ -4,7 +4,6 @@ import (
 	"flag"
 	"log"
 	"os"
-	"os/user"
 	"path/filepath"
 	"time"
 
@@ -108,20 +107,6 @@ func refreshAllWidgets() {
 	}
 }
 
-/* -------------------- Main -------------------- */
-
-var Config *config.Config
-var FocusTracker wtf.FocusTracker
-var Widgets []wtf.Wtfable
-var Pages *tview.Pages
-var MainPage *tview.Grid
-
-var (
-	commit  = "dev"
-	date    = "dev"
-	version = "dev"
-)
-
 func watchForConfigChanges(app *tview.Application, configFlag *string) {
 	w := watcher.New()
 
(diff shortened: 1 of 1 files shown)


wtf.go
```diff
@@ -104,51 +107,55 @@ func refreshAllWidgets() {
...
+	w := watcher.New()
+
+	// notify write events.
+	w.FilterOps(watcher.Write)
+
+	go func() {
+		for {
+			select {
+			case <-w.Event:
+				Pages.RemovePage("grid")
```
senorprogrammer: No need for this line, if you add a page with the same identifier as an existing one, it over-writes the existing one: https://github.com/rivo/tview/blob/74643a2db581c4268b25079d7a3f7a9ed441d0e2/pages.go#L52

(I didn't know that until now. Good stuff.)

wtf.go
```diff
@@ -104,51 +107,55 @@ func refreshAllWidgets() {
 	}
 }
 
+func watchForConfigChanges(app *tview.Application, configFlag *string) {
+	w := watcher.New()
```
senorprogrammer: I prefer not to use single-character variable names except as counters in loops. Otherwise I have to keep the reference in my head when working with the code. How about `watch`?

wtf.go
```diff
@@ -104,51 +107,55 @@ func refreshAllWidgets() {
...
+			case <-w.Closed:
+				return
+			}
+		}
+	}()
+
+	// Watch config file for changes.
+	if err := w.Add(*configFlag); err != nil {
+		log.Fatalln(err)
+	}
```
senorprogrammer: 👍

wtf.go
```diff
@@ -172,19 +179,65 @@ func main() {
...
+
+		wtf -help    | --help
+		wtf -version | --version
+	*/
+	homeDir, err := wtf.Home()
+	if err != nil {
+		os.Exit(1)
+	}
+
+	flagConf := flag.String("config", filepath.Join(homeDir, ".wtf", "config.yml"), "Path to config file")
```
senorprogrammer: `filepath.Join` - much better, very nice.

PR #66, opened 2018-05-29
Add notifications to Github widget

matrinox:
### Proposal
Add notifications list to GitHub widget. The final result looks like this:

screen shot 2018-05-28 at 8 26 22 pm

### Changes
- New GithubActivity model to contain notifications
- Github widget updated to display notifications after repo details


senorprogrammer:
This is a good idea. A couple of thoughts:

1. I'd like this behind a config flag for those who don't want it (on my home machine I use WTF to track personal repos no one else contributes to, don't need to see notifications for those).

2. What do the paging sigils in this case represent? The keyboard commands will switch between Github repos - how would I look at page 2 of the notifications? Perhaps instead of trying to show all notifications it could limit it to a user-configurable number (i.e.: only show the last ten, etc.)?

3. Take. a look at how Jira issues are displayed and color-coded for that widget. Would be great to have Github notifications follow a similar visual style.

senorprogrammer:
Checking in on this PR. If you don't want to continue it, let me know and I'll close it out.

matrinox:
@senorprogrammer Sorry for the long delay, busy with work. I've fixed conflicts and pushed it up. Can you take over? I implemented #1, which is optionally not showing notifications. If the user doesn't want notifications, just don't include `gitHub.notificationUser`; simple.

The other 2 are interesting but you may know it better. Just do the changes and I can learn from those. Thanks.

senorprogrammer:
Thanks @matrinox, I'll take a look.

senorprogrammer:
This seems to have merged cleanly. As soon as I have git repos that have notifications on them again, I'll finish this.

matrinox:
@senorprogrammer Just rebased so the commit history is clean. Checks have passed, please merge when you have time

senorprogrammer:
As is, I'm not a fan of this implementation of this feature for the following reasons:

* loading notifications _really_ slows down the loading of all GitHub data, and notifications are the part I'm not interested in
* adding `notificationUser` means I have my GitHub username in two config variables that do the same thing
* it'll take me a lot of work to get the look & feel of this consistent with the rest of the UI
* I don't have time to do that work vs. the rest of what needs to be done on WTF

@matrinox If you want to continue taking this forward, please do, with the above considerations in mind (the performance one is a deal-breaker).

senorprogrammer:
Closing as sbandoned

Issue #65, opened 2018-05-28
Be nice if the flags supported both long and short options


senorprogrammer:
i.e.: `wtf -v`, `wtf --version`.

Use https://github.com/jessevdk/go-flags ?

senorprogrammer:
Fixed by @jeangovil.

PR #68, opened 2018-05-29
now flags support both short and long names

jeangovil:
Issue: #65

### DIFF
--- a/wtf.go
+++ b/wtf.go
@@ -1,13 +1,13 @@
 package main
 
 import (
-	"flag"
 	"log"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Works well for `-v, --version`. For `-h, --help` it displays the help text but then runs through and launches the app. It looks like the module help has also been lost. `wtf --help clocks` should display the help text for the `Clocks` module, etc.

senorprogrammer:
I've fixed the issue with not halting on `help` and merged my branch in. It includes all the changes made in this PR. Nice work! 🎉

PR #95, opened 2018-05-31
Advanced Jira Search.

sherodtaylor:
added support for custom JQL

### DIFF
--- a/jira/client.go
+++ b/jira/client.go
@@ -7,11 +7,31 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
+	"strings"
 )
 
-func IssuesFor(username string) (*SearchResult, error) {
-	url := fmt.Sprintf("/rest/api/2/search?jql=assignee=%s", username)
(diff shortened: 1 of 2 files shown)


senorprogrammer:
This looks great. Could you add a comment with an example of how custom JQL would be used/added to config in this case so I can understand the outcome?

sherodtaylor:
So what this is doing is  creating a base query 
`assignee = "staylor" AND project = "Scrum"` 

Then if the param jql is inserted it would also add that like 
yaml
```
jql:  "priority in (Blocker, Critical)"
assignee: "staylor"
project: "Scrum"
```

would translate to
`assignee = "staylor" AND project = "Scrum" AND priority in (Blocker, Critical)`

Here's a good intro to JQL https://confluence.atlassian.com/jiracore/blog/2015/07/search-jira-like-a-boss-with-jql

senorprogrammer:
Thanks for the details. I'll give this a try against our Jira install.

senorprogrammer:
Thanks! Merged.

PR #96, opened 2018-05-31
Fixing the NewRelic deployments slice issue

lasanthak:
This fails with an unbound slice error when the revision value is smaller


lasanthak:
see https://github.com/senorprogrammer/wtf/pull/97

Issue #94, opened 2018-05-31
Linux Support


cryptmin:
The "Security" modules only support OSX. Please add support for Linux, or utilize a library that enables cross-platform compatibility. 

Love the design and idea behind the application! Great work!

interlock:
I'm about half done doing this right now. It's quite... Ubuntu specific, but that probably hits a good chunk of people.

nkoehring:
@interlock can you elaborate on "ubuntu specific"?

senorprogrammer:
Awesome. If someone puts up a PR, I'll merge it in.

interlock:
@nkoehring Sure can.

Wifi and DNS data comes from `nmcli` which I am not sure is commonly installed with most distros. 

Users is filtered from `who`, which is POSIX so :+1: for that

I could find no way to determine firewall state without `sudo` so put a a less assuming "NA" in it's place.

nkoehring:
@interlock nmcli is NetworkManager, common in most graphically rich distros. WTF is more focused on skilled users, so I would assume a good percentage isn't using NetworkManager (me included). There's iwconfig (old) and iwtools (current) for that which should be available on Ubuntu alike Systems too

interlock:
@nkoehring good to know. I updated the PR to use `iwgetid` for the network name, but I can't find any tools around getting the encryption. Any suggestions?

senorprogrammer:
Added with this commit: https://github.com/senorprogrammer/wtf/commit/a88a909e1060fbda54bdbbea6eb7166983aed419

Issue #87, opened 2018-05-31
Cannot `make install` unless already installed.


Sacro:
```
which wtf | xargs rm
rm: missing operand
Try 'rm --help' for more information.
Makefile:9: recipe for target 'install' failed
make: *** [install] Error 123
```

`which wtf` will always fail if it's not already installed.

jproffer:
you can get around this by just typing 'make' (without install) then 'go build'

Sacro:
I was just following the Quickstart on the front page!

Sacro:
`go build github.com/senorprogrammer/wtf: build output "wtf" already exists and is a directory`

senorprogrammer:
Pretty sure @Sacro is correct and that's a bug. In the meantime you can run `go install`. It won't populate the build values in the System module correctly, but it should install. If you want the version values you can then run `make install`.

senorprogrammer:
Fixed with this commit https://github.com/senorprogrammer/wtf/commit/a88a909e1060fbda54bdbbea6eb7166983aed419

rmbrntt:
As a heads up, still getting a similar issue when following the `make install` instructions with that latest fix commit
```
make install
which wtf | xargs rm || true
which: no wtf in (...:/usr/local/go/bin:...
rm: missing operand
```

fosspill:
Indeed, I'm experiencing this as well while following the guide

senorprogrammer:
@fosspill Which OS are you on?

fosspill:
Ubuntu 18.04

deltaxflux:
Yeah time to improve the installer.

PR #99, opened 2018-05-31
linux/ubuntu wraps for security

interlock:
Updates the `security` module to work with Linux. Uses runtime.GOOS to use platform specific methods.

Could use `// +build linux` to potentially only compile in funcs for each platform as well, but this is proof of concepts AFAIK.

Partly addresses #94


interlock:
Also addresses #87

PR #109, opened 2018-06-01
Feat: add an error message when WTF was failed

flw-cn:
This is just a minor change to avoid a small number of inexperienced
users not be intimidated by a minor mistake, such as the incorrect
setting of the TERM environment variable.

I hope this will be improved a bit of user experience for let WTF
more popular.

### DIFF
--- a/wtf.go
+++ b/wtf.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"time"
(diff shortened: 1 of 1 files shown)


senorprogrammer:
This looks like a great addition. What should I do to trigger it so I can see it execute in action?

evanfoster:
Speaking from experience, this causes the problem:

```bash
TERM=xterm-termite ./wtf
```

Glad I saw this! My `$TERM` var is set to `xterm-termite` and I was beating my head against the wall trying to get this thing to work. Doing this:

```shell
TERM=xterm-256color ./wtf
```
Made `wtf` start up. I don't know if this is in the docs (I looked and couldn't find it) but if it isn't you might want to mention acceptable values for `$TERM`.

senorprogrammer:
Great idea. Will add to docs at some point.

senorprogrammer:
This commit https://github.com/senorprogrammer/wtf/commit/7545eadaffe31e45052e1cd1097dd4988a78eced might be of interest to you. You can now specify the terminal type in your config file.

Issue #105, opened 2018-06-01
`sw_vers` doesn't work on Linux , how about `lsb_release -a` ?


XanthusL:
https://github.com/senorprogrammer/wtf/blob/bf4ab1056f8f50dd8269a2b639ed155c210e83b8/system/system_info.go#L21

XanthusL:
I don't think it works. Output of `lsb_release -a` has different keys with `sw_vers`.
Without tests on other distros, I'm not sure about which keys to use. But it is clear that we should use different keys for Linux.
```
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
```
```go
sysInfo := SystemInfo{
		ProductName:    m["ProductName"],
		ProductVersion: m["ProductVersion"],
		BuildVersion:   m["BuildVersion"],
	}
```

XanthusL:
Here is the output on CentOS, I think it's OK to use `Distributor ID` as `ProductName`, `Description` as  `ProductVersion` and `Release` as   `BuildVersion`. #124 
```
LSB Version:    :core-4.1-amd64:core-4.1-noarch
Distributor ID: CentOS
Description:    CentOS Linux release 7.4.1708 (Core) 
Release:        7.4.1708
Codename:       Core
```

PR #110, opened 2018-06-01
Quick fix for #105

deltaxflux:
- add lsb_release because sw_vers doesn't work on linux 
- add system check


senorprogrammer:
Thanks for making this. I had to add a couple tweaks to get it compile; it's in master now.

deltaxflux:
yeah sorry I forget to add the runtime lib

PR #118, opened 2018-06-01
strip space and ignore empty users from who

### LINKED ISSUE #117: Singular user appended ','
@interlock Thank you for your work on the WTF Linux Compatability. 

When there is only a single user session a comma is appended to the username.
2018-06-01-132212_411x295_scrot

Consider using `users` instead of `who -us`. Both are GNU CoreUtils.

interlock:
Remove spaces and ignore zero length lines.

closes #117


senorprogrammer:
Thanks!

PR #120, opened 2018-06-01
Moving git recent commits above changed files

NickBusey:
When you have many changed files, recent commits can be pushed outside of the screen.

Since only the changed files list takes a config parameter to control how many are displayed, I thought it made more sense to have it on top.

### DIFF
--- a/git/display.go
+++ b/git/display.go
@@ -24,9 +24,9 @@ func (widget *Widget) display() {
 	str = str + " [red]Branch[white]\n"
 	str = str + fmt.Sprintf(" %s", repoData.Branch)
 	str = str + "\n"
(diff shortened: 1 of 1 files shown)


senorprogrammer:
I'm going to reject this one for three reasons:

1. "Changed Files" shows what I'm currently doing, "Recent Commits" is what was done. What I'm currently doing is more important to me, at a glance, than what was done.
2. If you need more room, the widget can probably be sized larger onscreen.
3. Most importantly, for me a large number of changed files is often an indicator of a commit out of control. Small changesets are usually better than large changesets. I have done a few large enough to have the effect you describe, and the effect serves as a warning about the size of the change.

NickBusey:
Lol k.

PR #123, opened 2018-06-02
Fix "When deleting a todo item, select the next item up if there is one"

stegmanh:
This appears to be an issue when deleting the last item in a list in the todo module. This simply implements a check to decrement the selected if a user deletes the last item in a list, else, it behaves normally. 

Fixes issue: https://github.com/senorprogrammer/wtf/issues/101

Thanks, and let me know if you have any feedback!

### DIFF
--- a/todo/list.go
+++ b/todo/list.go
@@ -33,6 +33,9 @@ func (list *List) CheckedItems() []*Item {
 
 func (list *List) Delete() {
 	list.Items = append(list.Items[:list.selected], list.Items[list.selected+1:]...)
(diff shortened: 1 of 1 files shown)


senorprogrammer:
This is great, thank you. Unfortunately the merge of your branch with master had a few merge conflicts so rather than resolve those I've just copied your lines in this commit https://github.com/senorprogrammer/wtf/commit/4c620a9c3cec19279402146570af50021bc10df0 rather than fold in this PR.

jeangovil:
All you need is:
`list.Prev()`

PR #127, opened 2018-06-02
added cryptolive widget

jeangovil:
A widget to compare cryptocurrencies using [cryptocompare.com](https://cryptocompare.com)
screenshot from 2018-06-02 14-14-22


jeangovil:
@senorprogrammer no comments? If you think this widget is not beneficial or you don't like it, I'm cool if you close this PR.

senorprogrammer:
I like this addition a lot, I've just been thinking about how to implement `n` cryptocurrency exchanges rather than just this one, at the structure level. Currently thinking it should be:
```
wtf/cryptoexchanges/bittrex
wtf/cryptoexchanges/coinbase/
wtf/cryptoexchanges/cryptolive/
```
etc. rather than having them all be individual at the top level. Thoughts?

jeangovil:
I think we can make a general widget named `cryptoexchange` and a simple web server to handle requests. This way we will avoid making the binary file larger so we can add support for any API we want.

senorprogrammer:
Could you elaborate more? I don't thin I fully understand the data flow and dependencies.

jeangovil:
Forget about it. I will close this PR and make another for Bittrex and Cryptolive.

senorprogrammer:
Wait, what happened? I do want to merge this one in, I think this is the right direction to go. Mind if I reopen? I just haven't had time to do anything on this today.

senorprogrammer:
As for the web sockets idea, I'm not going to tackle that for two reasons:

1. That sounds like a complete and excellent project in its own right. No small amount of work to do that. As a service, I can see a lot of different things tying into it.
2. I think it can live nicely in parallel with dedicated modules. If you built that service and the module that used it, I would definitely include that in. And I think there's room for vendor-specific ones for Bittrex, CryptoLive, etc. I'm not yet worried about the binary size.

senorprogrammer:
Thanks @jeangovil.

cryptoexchanges/cryptolive/cryptolive.go
```diff
@@ -15,33 +15,7 @@ type toCurrency struct {
...
-	LSK  float32 `json:"LSK"`
-	ELA  float32 `json:"ELA"`
-	DTA  float32 `json:"DTA"`
-	NANO float32 `json:"NANO"`
-	WTC  float32 `json:"WTC"`
-	DOGE float32 `json:"DOGE"`
-	USD  float32 `json:"USD"`
-	EUR  float32 `json:"EUR"`
-}
+type cResponse map[string]float32
```
senorprogrammer: Beautiful.

PR #131, opened 2018-06-02
Changes in git plugin - added checkout and pull

BlackWebWolf:
While git plugin itself is great, adding this two functionalities will greatly improve working speed - as I'm a lazy one, finding all repos in one place in terminal is great - but without pull and checkout it is a bit limited. Adding commit and push would be a bit too much in my opinion, but this will help with quick checking changes.

### DIFF
--- a/git/git_repo.go
+++ b/git/git_repo.go
@@ -69,6 +69,19 @@ func (repo *GitRepo) repository() string {
 
 	return str
 }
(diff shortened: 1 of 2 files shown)


senorprogrammer:
This looks like a great addition. I'll try to test this out this weekend.

senorprogrammer:
Good stuff, thanks!

PR #133, opened 2018-06-02
Minor template changes

deltaxflux:
I bump the version number to 0.04 and I add the pull request subfolder. Nothing special.


deltaxflux:
I was wondering why using 0.0.4 instead of 0.04?

senorprogrammer:
Somewhat following semver (https://semver.org). Technically it should probably be 0.4.0 however that version number seems to grandiose for where it's at, esp. given the problems many people are having with installation, dependency management, etc. It's very much in a "works for me" stage at the moment so even though major features are being added, I don't feel it's at a v0.1.0 stage yet.

PR #138, opened 2018-06-03
use widget.View.SetText to print text

jeangovil:
Now if  the `tview` library ever changes the functionality of printing text, it will work fine. And also it doesn't make sense to write the code that has already written.


deltaxflux:
Good point :)

senorprogrammer:
My favourite kind of PR: one that objectively makes existing code better. Much appreciated.

PR #139, opened 2018-06-03
Banish .DS_Store

askl56:
I see it's in the gitignore already but just removing the pesky DS_Store files.


senorprogrammer:
Thanks!

PR #141, opened 2018-06-03
Made System and Build Info Mode awesome again

### LINKED ISSUE #140: Make Build/System mod awesome again
### Feature Request

|    Q        |   A
|------------ | ------
| New Feature | Kinda
| RFC         | yes
| BC Break    | No
#### Summary

The Build/System mod is available in the default config, hence should be awesome. My current Status:

spectacle o28816

As the name implies, Maybe some cool system information with the build time/version would be cool :call_me_hand:

FarhadF:
Closes #140 

Added various system information to the widget available at the following screenshot:

spectacle tj3791

PS. I could only test this on linux, if you got a mac, give it a whirl.

### DIFF
--- a/system/system_info.go
+++ b/system/system_info.go
@@ -1,67 +1,32 @@
 package system
 
 import (
-	"os/exec"
-	"runtime"
-	"strings"
-
-	"github.com/senorprogrammer/wtf/wtf"
+	"github.com/matishsiao/goInfo"
 )
 
+//systemInfo will hold system information using goInfo pkg
 type SystemInfo struct {
-	ProductName    string
-	ProductVersion string
-	BuildVersion   string
+	GoOs     string
+	Kernel   string
+	Version  string
+	Platform string
+	OS       string
+	Hostname string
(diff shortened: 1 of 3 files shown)


senorprogrammer:
The mission statement for WTF is "rapidly viewing important at-a-glance information easily".  This might be technically more accurate as it provides a lot more detail, but it doesn't pass that mission statement test. For one,  it no longer includes the OS information that I actually do need. On macOS, knowing that my underlying kernel version is `16.7.0` is not something I've ever wanted. Knowing that my macOS version is `10.12.6`, and that the build of _that_ is `16G1314` is.

The solution is not to add even more fields, but rather less, only the ones that someone might actually want day-to-day or week-to-week, and mercilessly cutting out the others (for instance, `Cores`).

senorprogrammer:
That's all a long way of saying I think this needs multiple modules. The existing `System` one, and another, perhaps called `OS`, that provides the deeper, more esoteric details.

PR #147, opened 2018-06-04
added Bittrex module

jeangovil:
Added Bittrex module. It only supports market summary just for now.
screenshot from 2018-06-04 06-10-00


senorprogrammer:
This is great, and thank you for adding the docs as markdown. Excellent.

PR #148, opened 2018-06-04
Bug fix: lsb_release, Improvement: Power mod for Linux

XanthusL:
```go
// cmd = exec.Command("lsb_release -a", arg...)
// exec: "lsb_release -a": executable file not found in $PATH

// cmd = exec.Command("lsb_release", "-a")
```
Just put `-a` in `arg`, tested on Arch Linux

`pmset` doesn't work on Linux, `upower` works, tested on Arch Linux

### DIFF
--- a/power/battery_linux.go
+++ b/power/battery_linux.go
@@ -23,13 +23,7 @@ type Battery struct {
 }
 
 func NewBattery() *Battery {
-	var battery *Battery
-	//sh := `upower -i $(upower -e | grep '/battery') | grep --color=never -E "state|to\ full|to\ empty|percentage"`
-	battery = &Battery{
-		//args: []string{"-c", sh},
-		//cmd:  "/bin/sh",
-	}
-	return battery
+	return &Battery{}
 }
 
 /* -------------------- Exported Functions -------------------- */
--- a/power/battery_linux.go
+++ b/power/battery_linux.go
@@ -1,18 +1,17 @@
-// +build !linux
+// +build linux
 
 package power
 
 import (
 	"fmt"
 	"os/exec"
-	"regexp"
 	"strconv"
 	"strings"
 
 	"github.com/senorprogrammer/wtf/wtf"
 )
 
-const TimeRegExp = "^(?:\\d|[01]\\d|2[0-3]):[0-5]\\d"
+var batteryState string
 
 type Battery struct {
 	args   []string
@@ -24,12 +23,13 @@ type Battery struct {
 }
 
 func NewBattery() *Battery {
-	battery := Battery{
-		args: []string{"-g", "batt"},
-		cmd:  "pmset",
+	var battery *Battery
+	//sh := `upower -i $(upower -e | grep '/battery') | grep --color=never -E "state|to\ full|to\ empty|percentage"`
+	battery = &Battery{
+		//args: []string{"-c", sh},
+		//cmd:  "/bin/sh",
 	}
-
-	return &battery
+	return battery
 }
 
 /* -------------------- Exported Functions -------------------- */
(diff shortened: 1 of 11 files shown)


XanthusL:
screenshot from 2018-06-04 14-40-39

power/battery_linux.go
```diff
@@ -0,0 +1,118 @@
...
+	Charge    string
+	Remaining string
+}
+
+func NewBattery() *Battery {
+	var battery *Battery
+	//sh := `upower -i $(upower -e | grep '/battery') | grep --color=never -E "state|to\ full|to\ empty|percentage"`
+	battery = &Battery{
+		//args: []string{"-c", sh},
+		//cmd:  "/bin/sh",
```
senorprogrammer: Please remove this commented-out code?

power/battery.go
```diff
@@ -1,3 +1,5 @@
+// +build !linux
```
senorprogrammer: So I understand: this tells the compiler not to compile this module on Linux....

power/battery_linux.go
```diff
@@ -0,0 +1,118 @@
+// +build linux
```
senorprogrammer: ...and instead compile this module on Linux, which has exactly the same names, but different implementations? If so, that's very sweet.

system/system_info.go
```diff
@@ -22,7 +22,8 @@ func NewSystemInfo() *SystemInfo {
 	var cmd *exec.Cmd
 	switch runtime.GOOS {
 	case "linux":
-		cmd = exec.Command("uname -a", arg...)
+		arg = append(arg, "-a")
+		cmd = exec.Command("lsb_release", arg...)
```
senorprogrammer: For my own edification as a non-Linux user, could you explain what the different here is?

XanthusL:
```sh
$uname -a
Linux Quantum 4.16.13-1-ARCH #1 SMP PREEMPT Thu May 31 23:29:29 UTC 2018 x86_64 GNU/Linux
```
Information from `uname` is not enough. There is nothing about the distro, such as Ubuntu, Fedora, etc. It just shows `GNU/Linux`.

But `lsb_realease -a` outputs details of a distro.
```sh
$lsb_release -a
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
```

power/battery_linux.go
XanthusL: At first, `switch runtime.GOOS`, but there were many switch-case clauses. So I created files for Linux

system/system_info.go
senorprogrammer: Great, thanks.

deltaxflux: don't agree on that

senorprogrammer: @deltaxflux Can you elaborate? I'm not a Linux guy, this is outside my domain.

XanthusL: @deltaxflux Did you mean `lsb_release` not installed on some systems? 
How about  reading ` /etc/lsb-release`
But I don't know if /etc/lsb-release exists without lsb_release installed.

deltaxflux: I mean that lsb_release is not a core package of linux therefore we should use uname because it is shipped with Linux. I don't know if /etc/lsb-release exists if it is not installed but it doesn't make any sense, do you think so?

senorprogrammer:
I'm going to leave this one to you guys to sort out. Let me know the outcome.

system/system_info.go
XanthusL: uname is a core package but we cant get necessary info from it. 

Most info from uname can be got with std lib of golang, such as runtime.GOOS, runtime.GOARCH, os.Hostname()

I think its worth to have a try on lsb_release, and `screenfetch` is using lsb_release too

deltaxflux: Yeah agree

senorprogrammer:
Great work, thank you guys.

deltaxflux:
Was seriously bullshit what was done there.

senorprogrammer:
What do you mean? From the conversation above it looks to me like you two agreed that this PR was the way to go. Not so? There's something here I'm not understanding.

deltaxflux:
No the first problem is that we have like 4 issues open with the same discussion and I told already that  uname is better. there's no point using lsb-release whatsoever

senorprogrammer:
Above, @XanthusL wrote: "I think its worth to have a try on lsb_release"
You then wrote: "Yeah agree"

I took that to mean you two agreed on lsb_release - not the case? I'm a bit confused now.

deltaxflux:
Oh well your right. Sorry for the confusion. I agreed with the fist part.

XanthusL:
- uname is always available but we cant get all info from it.
- lsb_release provides all info but maybe not installed on some systems.

So, maybe we should dive into lsb_release and implement it instead of using it directly. I think this is the best way, but there would be more work to do.

senorprogrammer:
How about checking to see if its installed? If it is, use `lsb_release`, if not then fall back to what `uname` can provide.

deltaxflux:
yeah thats not a problem.

deltaxflux:
**Like this:**
`if ! hash lsb_release 2>/dev/null;then uname;else lsb_release;fi`

Issue #89, opened 2018-05-31
Gracefully handle no specified DNS settings


senorprogrammer:
If the current network interface has no custom DNS settings, it displays "There aren't any DNS Settings..." that writes off the screen. Handle this case gracefully.

senorprogrammer:
Closed by https://github.com/senorprogrammer/wtf/pull/151

PR #151, opened 2018-06-04
Fix issue #89

lixin9311:
Better handle dns on macOS

use the command 
`scutil --dns | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}'`
to get dns on macOS

### DIFF
--- a/security/dns.go
+++ b/security/dns.go
@@ -42,7 +42,7 @@ func dnsLinux() []string {
 }
 
 func dnsMacOS() []string {
-	cmdString := `scutil --dns | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}'`
+	cmdString := `scutil --dns | head -n 7 | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}'`
 	cmd := exec.Command("sh", "-c", cmdString)
 	out := wtf.ExecuteCommand(cmd)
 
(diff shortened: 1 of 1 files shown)


security/dns.go
```diff
@@ -42,7 +42,8 @@ func dnsLinux() []string {
 }
 
 func dnsMacOS() []string {
-	cmd := exec.Command("networksetup", "-getdnsservers", "Wi-Fi")
+	cmdString := `scutil --dns | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}'`
+	cmd := exec.Command("sh", "-c", cmdString)
```
senorprogrammer: Interesting approach. Returns duplicate entries on my machine:
```bash
~/go/src/github.com/senorprogrammer/wtf master
❯ scutil --dns | grep -o '[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}\.[0-9]\{1,3\}'
1.1.1.1
1.0.0.1
1.1.1.1
1.0.0.1
```

DominicGunn: Could maybe fix that with a `sort -u`?

lixin9311:
I don't know the network stack of macOS very much. The duplicated ones are
for the scoped network. Anyway, you can use head to keep the first n lines
of the output of scutil, then pipe the result to grep.

security/dns.go
lixin9311: fixed. it may be not so elegant.

https://superuser.com/questions/734063/what-is-a-scoped-dns-query
about the scoped queried. I don't know what setting will break the code, but it looks like working well for me.

PR #160, opened 2018-06-04
System: Correct command call for linux

patrickelectric:
Signed-off-by: Patrick José Pereira <patrickelectric@gmail.com>

### DIFF
--- a/system/system_info.go
+++ b/system/system_info.go
@@ -22,7 +22,8 @@ func NewSystemInfo() *SystemInfo {
 	var cmd *exec.Cmd
 	switch runtime.GOOS {
 	case "linux":
-		cmd = exec.Command("uname -a", arg...)
+		arg = append(arg, "-a")
+		cmd = exec.Command("lsb_release", arg...)
 	case "darwin":
 		cmd = exec.Command("sw_vers", arg...)
 	default:


patrickelectric:
uname does not return all necessary info requested [here](https://github.com/senorprogrammer/wtf/blob/master/system/system_info.go#L46).
```sh
$ Uname -a 
Linux dell 4.16.13-1-ARCH #1 SMP PREEMPT Thu May 31 23:29:29 UTC 2018 x86_64 GNU/Linux
```
```sh
$ lsb_release -a                                          
LSB Version:	1.4
Distributor ID:	Arch
Description:	Arch Linux
Release:	rolling
Codename:	n/a
```

system/system_info.go
```diff
@@ -22,7 +22,8 @@ func NewSystemInfo() *SystemInfo {
 	var cmd *exec.Cmd
 	switch runtime.GOOS {
 	case "linux":
-		cmd = exec.Command("uname -a", arg...)
+		arg = append(arg, "-a")
+		cmd = exec.Command("lsb_release", arg...)
```
senorprogrammer: Looks to me like this is handled in this commit: https://github.com/senorprogrammer/wtf/pull/148/files
in PR 148, correct?

patrickelectric: Yep, you can close this PR if necessary.

senorprogrammer:
Thanks

system/system_info.go
deltaxflux: uname -a is better because it is a core package of linux.

PR #172, opened 2018-06-05
Built Graph Widget, working with basic sample

BillKeenan:
Hi,

I made a simple bargraph widget on top of the existing text widget. 

Looks pretty nice, work to go to make it 'fancy'

screenshot 2018-06-05 16 06 40


bargraph/widget.go
```diff
@@ -0,0 +1,95 @@
...
+		return
+	}
+
+	if started == false {
+		// this code should run once
+		go func() {
+			for {
+				time.Sleep(time.Duration(widget.updateInterval) * time.Second)
+			}
+		}()
```
senorprogrammer: What's the purpose of this go routine? This just seems to sleep it over and over again, with no external  effect. What am I missing?

bargraph/widget.go
```diff
@@ -0,0 +1,95 @@
...
+	var stats [lineCount][2]int64
+
+	for i := lineCount - 1; i >= 0; i-- {
+
+		stats[i][1] = time.Now().AddDate(0, 0, i*-1).Unix() * 1000
+		stats[i][0] = int64(rand.Intn(120-5) + 5)
+
+	}
+
+	widget.GraphWidget.BuildBars(20, "🌟", stats[:])
```
senorprogrammer: Where does the data come from for this graph? It looks like it's just building random lines. Or is this a demo for `GraphWidget`, below?

wtf/graph_widget.go
```diff
@@ -0,0 +1,160 @@
...
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/rivo/tview"
+)
+
+type GraphWidget struct {
```
senorprogrammer: I think this should be renamed to not include `Widget`. In WTF a directory of executable files that comprise a single service are a "module", and the part that is effectively `main` that manages the content that will be displayed onscreen, is called a "widget".  Might be confusing (for instance, above, `bargraph/widget/Widget` is including `GraphWidget`).

How about calling this just `BarGraph`?

bargraph/widget.go
BillKeenan: correct, that's holdover from the template code, i'll remove that.

bargraph/widget.go
BillKeenan: yes this is a demo widget

wtf/graph_widget.go
BillKeenan: sounds good.

bargraph/widget.go
BillKeenan: I'll comment as such.

BillKeenan:
ok, those items have been addressed here.

PR #186, opened 2018-06-08
create another module to use another ipinfo api

Fengyalv:
The original ip API has a rate limit for 1000 per day, which is a little bit easy to reach.

This module use an API from `http://ip-api.com`, which has a higher rate limit for 120 per minute.

This API also provide more informations about ip such as `AS`, `timezone`.

### DIFF
--- a/ipinfohigherlimit/widget.go
+++ b/ipinfohigherlimit/widget.go
@@ -1,0 +1,141 @@
+package ipinfohigherlimit
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+	"text/template"
+
+	"bytes"
+
+	"github.com/olebedev/config"
+	"github.com/senorprogrammer/wtf/wtf"
+)
+
+// Config is a pointer to the global config object
+var Config *config.Config
+
+// Widget widget struct
+type Widget struct {
+	wtf.TextWidget
+	result string
+	colors struct {
+		name, value string
+	}
+}
+
+type ipinfo struct {
+	Query        string  `json:"query"`
+	ISP          string  `json:"isp"`
+	AS           string  `json:"as"`
+	City         string  `json:"city"`
+	Region       string  `json:"region"`
+	Country      string  `json:"country"`
+	CountryCode  string  `json:"countryCode"`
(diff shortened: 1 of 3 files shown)


ipinfohigherlimit/widget.go
```diff
@@ -0,0 +1,141 @@
...
+
+// Refresh refresh the module
+func (widget *Widget) Refresh() {
+	if widget.Disabled() {
+		return
+	}
+
+	widget.UpdateRefreshedAt()
+	widget.ipinfo()
+	widget.View.Clear()
```
jeangovil: Don't need this when use `widget.View.SetText()` to print text.

ipinfohigherlimit/widget.go
```diff
@@ -0,0 +1,140 @@
...
+	widget.View.SetWrap(false)
+
+	widget.config()
+
+	return &widget
+}
+
+// Refresh refresh the module
+func (widget *Widget) Refresh() {
+	if widget.Disabled() {
```
jeangovil:
```
if widget.Disabled() {
	return
}
```
It's not necessary any more.

ipinfohigherlimit/widget.go
```diff
@@ -0,0 +1,135 @@
...
+		return
+	}
+	req.Header.Set("User-Agent", "curl")
+	response, err := client.Do(req)
+	if err != nil {
+		widget.result = fmt.Sprintf("%s", err.Error())
+		return
+	}
+	defer response.Body.Close()
+	if err != nil {
```
jeangovil:
```
if err != nil {
	widget.result = fmt.Sprintf("%s", err.Error())
	return
}
```
Sorry I can't see the reason for checking the same `err` two times(lines 73-76 & 78-81)

Fengyalv: Sorry, this part is from the original ipinfo. So I didn't change it. I will correct it.

senorprogrammer:
I like the module, I don't like the name. `ipapi` seems like a reasonable name for it, given the service it uses.

senorprogrammer:
Thanks!

PR #191, opened 2018-06-08
Allow use of project list in Jira module config

baustinanki:
For my own use case (and anyone in a similar situation), the Jira
module is more useful if I can specify a list of projects to
display issues from, rather than no project (which selects all
projects) or one specific project.

New supported syntax:
`  project: ["PROJA", "PROJB"]`

If this is merged, documentation for the Jira module should be
updated accordingly. (Sorry, I would have done this myself but
I'm not sure what the right place is - when grepping I find
multiple places in the repo containing documentation strings
and I'm not sure what the "master" location is)

Specifying a single project (or no project) is still supported -
behavior shouldn't change for anyone who doesn't change their
config.


senorprogrammer:
This is a great PR - thanks!

PR #192, opened 2018-06-08
Add Gitlab module

dlom:
This module is heavily based on the existing github module.  The underlying logic is somewhat different because of differences in github's and gitlab's APIs.  The user interface is largely the same, except that all github terminology has been converted to gitlab terminology.

The actual code changes occur in the first commit, and the second commit only has docs changes.


senorprogrammer:
Thanks - I'll try and get this merged in today. I don't use Gitlab. Would you be able to attach a screenshot?

senorprogrammer:
Thanks for the PR. I'm just finishing up up release 0.0.7 now, will try to get this into 0.0.8.

PR #193, opened 2018-06-08
Fix newline in git module repo names breaking display

baustinanki:
After setting up the git module with multiple repos and switching
between them, I observed some graphical wonkiness in the display:

After adding some log statements, I tracked it down to the
`GitRepo.Repository` field having a newline in it after it's set
from a command execution's stdout. This change strips the
repository path of spaces when assigning to the `Repository` field,
which fixes the display issues.


senorprogrammer:
Thanks for the PR - will merge once TravisCI passes.

PR #195, opened 2018-06-09
Add circleci module

mcordell:
Hi,

Thanks for the great program. This PR adds support for [CircleCI](https://circleci.com/account/api).  I patterned the module off of the JIRA module. As such, it needs a Circle API token, which can be found at https://circleci.com/account/api. This is passed under the environmental variable WTF_CIRCLE_API_KEY.

The only other note is that I cribbed heavily from [go-circleci](https://github.com/jszwedko/go-circleci) in order to make the client, rather than introduce a new dependency.  The client could be replaced with that library if you are comfortable introducing a new dependency.

### DIFF
--- a/circleci/build.go
+++ b/circleci/build.go
@@ -1,0 +1,10 @@
+package circleci
+
+type Build struct {
+	AuthorEmail        string            `json:"author_email"`
+	AuthorName         string            `json:"author_name"`
+	Branch             string            `json:"branch"`
+	BuildNum           int               `json:"build_num"`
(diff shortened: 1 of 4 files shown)


senorprogrammer:
Awesome - I'll try to get this merged in today. I don't use CircleCI. Would you be able to attach a screenshot?

mcordell:
screen_shot_2018-06-09_at_9_16_37_am
two

senorprogrammer:
Thanks!

PR #198, opened 2018-06-10
Blockfolio widget

andrewzolotukhin:
Here is blockfolio widget which I use to display my crypto holdings and 24 hours percent changes of these holdings.
There is an option if it should display values in dollars (or just amount of coins + 24h percentage change)
How to get device_token for api: https://github.com/bob6664569/blockfolio-api-client
blockfolio


jeangovil:
Jesus. Why changed the packages' name?

andrewzolotukhin:
@jeangovil in which file? I was changing it while developing, this is because of my local environment issues. Bit in pull request names are correct.

andrewzolotukhin:
Sorry, you're right )) cancelling it

PR #211, opened 2018-06-16
Google Calendar: multiple calendars, all-day event date markings

lsipii:
The commit alters the gcal-modules events fetching so that:

- Pre-commit: 
1. The gcal Fetch()-method gathered only the users "primary" events,  so for example I had no work-related events listed as those are in a separate calendar ID
2. The widget.. eventTimestamp()-method always displayed the time as "first of January" for the events that had no time, the "all-day" events, on the calendar. 

- Post-commit: 
1. The gcal Fetch()-method gathers all the calendar IDs the user has a write-access on, then combines the resulting calendar events in an array and sorts by the calendar times. 
2. Added a simple correcting if-clause when displaying the timeless-day-events. According to the Google calendar API-docs the calendar.Event.Start.Date -string has a length only when the event is a "all-day" event. (Does not clearly say so but confirmed by trying with the hope of gaining some extra experience points).

This is somewhat related to the issue: https://github.com/senorprogrammer/wtf/issues/47

Issues:

- The commit needs some serious audition and eyeing applied: the committers first time touch with Golang. Had to google how to define variables.. how to loop.. what's what. Please regard the commit as a suggestion and then rewrite, figure it out nicely!

- The gcal.eventCount setting is applied to all the calendars

### DIFF
--- a/gcal/client.go
+++ b/gcal/client.go
@@ -17,6 +17,7 @@ import (
 	"os/user"
 	"path/filepath"
 	"time"
+	"sort"
 
 	"github.com/senorprogrammer/wtf/wtf"
 	"golang.org/x/oauth2"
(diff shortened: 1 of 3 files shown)


senorprogrammer:
Thanks for the work! I'll take a look.

BTaskaya:
LGTM

senorprogrammer:
Great work, thanks! Merged.

PR #214, opened 2018-06-19
Update GCal module's "time until event" text more frequently

baustinanki:
While it's reasonable to have a refresh interval of several minutes
between the GCal module hitting the server, the fact that the module
(helpfully) displays the time until an event means that as times
approach zero, they become relatively inaccurate - "3m" might show
up next to a meeting that you were supposed to be in 2 minutes ago.
(don't worry, that hasn't happened to me yet)

This change introduces a goroutine to the GCal module that will
update the module text using the most recently cached copy of
events. By default, it runs every 30 seconds (this could be changed
to 0 to make it opt-in if desired) or however many seconds is
specified by the new config option `wtf.mods.gcal.textInterval`.

To make sure the goroutine doesn't cause any synchronization issues
if a text update were to trigger at the same time as a module
refresh, this change also adds the use of `sync.Mutex` to gate
setting GCal's text. Since both mutexes and having a goroutine within
individual modules are not the sort of thing any other code in
`wtf` does at this point in time, I'm definitely open to feedback
on whether there might be better ways to do this.

Additionally, the change in `wtf.go` fixes an issue I noticed when
testing if my `Disable()` function worked - the main `Widgets`
array wasn't being cleared out after disabling all widgets before
re-creating new copies of them, so it still held pointers to the
old ones.

### DIFF
--- a/gcal/widget.go
+++ b/gcal/widget.go
@@ -49,6 +49,10 @@ func (widget *Widget) Disable() {
 /* -------------------- Unexported Functions -------------------- */
 
 func (widget *Widget) display() {
+	if widget.events == nil || len(widget.events.Items) == 0 {
+		return
+	}
+
 	widget.mutex.Lock()
 	defer widget.mutex.Unlock()
 	widget.View.SetText(fmt.Sprintf("%s", widget.contentFrom(widget.events)))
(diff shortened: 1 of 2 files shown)


senorprogrammer:
This is a feature I've long wanted. I will try to get this in today. Thanks!

gcal/widget.go
```diff
@@ -254,3 +274,24 @@ func (widget *Widget) until(event *calendar.Event) string {
...
+	defer tick.Stop()
+outer:
+	for {
+		select {
+		case <-tick.C:
+			if widget.events != nil && len(widget.events.Items) > 0 {
+				widget.display()
+			}
+		case <-widget.ch:
+			break outer
```
senorprogrammer: Nice! I learned something about Go here.

PR #215, opened 2018-06-19
Fix documentation for logging module

anandsudhir:
https://github.com/senorprogrammer/wtf/commit/6acf1775b84bfd07cc272a0619af822c93385cac 
changed the name of the module but inconsistencies still remain in the documentation and in the widget names.

Fix inconsistencies  in naming and documentation for the logging/logger module

#### Is this related to an existing Issue? If so, which one?
No open issues on this

### DIFF
--- a/_site/content/posts/modules/logger.md
+++ b/_site/content/posts/modules/logger.md
@@ -1,5 +1,5 @@
 ---
-title: "Logging"
+title: "Logger"
 date: 2018-06-16T14:22:18-07:00
 draft: false
 ---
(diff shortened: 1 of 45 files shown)


anandsudhir:
Ideally, the name of the widget should be changed from logging to logger too.

anandsudhir:
I will submit a patch to this with the change in widget name later today

BillKeenan:
I tried out the new logging module, works well.

Would be good to mention that if enabled=false in the config, you can still log, it just doesn't display the widget.

senorprogrammer:
Thanks!

PR #220, opened 2018-06-22
[CircleCi] - Green build color for fixed builds

jdsingh:
If build is successful after a failed build, the status returned by Circleci API is `fixed` instead of `success`.

So it makes sense to show green color build status in that scenario as well.


senorprogrammer:
Thanks for this!

Issue #223, opened 2018-06-22
Proposal: Re-work format of calendar view


baustinanki:
#### What problem does this solve?

For a set of mocked-up fake events, here's what the current calendar view looks like:

![](https://i.imgur.com/995yEzJ.png)

There's nothing particularly wrong with it, but there are a couple ways I think it could be improved.

The most important aspect of an event - when it takes place - is somewhat tucked away, towards the end of the third line of an event. If you want to tell from just a glance when you're next going be pulled away from your desk, it's not the most obvious place to look.

Another important aspect of calendars is getting a sense of, from looking at it, "what is my day like today?" The grouping of events (with only an extra blank line separating one day from the next) doesn't give a lot of visual encapsulation of what makes up your "day".

#### How do envision it working?

I propose a format somewhat like the following (feedback/suggestions welcome):

![](https://i.imgur.com/drtf5iX.png)

- Centered date across the top makes it clear which block of space constitutes each day
- Time of each event is front and center at the beginning of it, also makes events look "sorted"
- By having the date centered at the start of each day, we don't have to repeat it for each event, allowing us to consolidate information to 2 lines instead of 3, leaving room for more events in the same amount of space

Depending on what people think, this could be an option (we could have some predefined formats that you select from in the config), or we could go full-on git-pretty-formatting with % specifiers and stuff to allow full customization (that's obviously a more complicated change). But, in general, I've been using this format locally for the last week or so and I really like it.

senorprogrammer:
When you say you've been using it locally, does that mean you could have a PR to commit that back against master? Because I like the look of what you've done a lot. Makes a lot of sense to me.

baustinanki:
>does that mean you could have a PR to commit that back against master?
There are a couple pieces I should clean up first but yeah, I could have a PR of that pretty soon.

senorprogrammer:
If you're playing along at home, the PR for this is here: https://github.com/senorprogrammer/wtf/pull/224

senorprogrammer:
Closed with this PR https://github.com/senorprogrammer/wtf/pull/224

PR #224, opened 2018-06-22
Re-work calendar view

baustinanki:
As discussed in issue #223 (which has pictures that better explain
the changes here), this change:
- Removes the date from individual events, instead centering a
  title at the start of each day with the date (which uses a new
  configurable color, `wtf.mods.gcal.colors.day`)
- Consolidates 3 lines per event down to 2, moving timestamp to
  front of each event
- Makes the time-until-event text turn red when under 30 minutes
  (wasn't discussed in the issue but was another thing I added
  locally for this, feel free to discard if unwanted)

New format is:
```
           Monday, Jun 25
 x 13:00 Super Cool Meeting Title 2h
 Event location

 x 14:00 Also Super Cool Meeting 3h
 Event location

          Tuesday, Jun 26
 ...
```


gcal/widget.go
```diff
@@ -158,7 +173,7 @@ func (widget *Widget) eventTimestamp(event *calendar.Event) string {
 		return startTime.Format(wtf.FriendlyDateFormat)
```
baustinanki: One caveat - I'm not sure what change we should make for this branch of the conditional. I suspect we don't want the current behavior (there's already a date heading, so printing the date is redundant, and now won't fit as well on the same line as the title), but this branch doesn't actually trigger for any events in my calendar so I'm not sure what the use case is (all-day events, maybe?) or what this _should_ look like in the new format.

senorprogrammer: Pretty sure that's for all-day events. I'll try this out and see how it looks with those.

senorprogrammer:
Yeah, it does some strange thing when `multiCalendar` is enabled. I'll try to sort that out in the next couple days (single-calendar view looks awesome).

PR #226, opened 2018-06-23
added cryptocompare toplist support to cryptolive

jeangovil:
**If you don't want to see toplist just remove it from cryptolive config**


senorprogrammer:
Nice work.

PR #227, opened 2018-06-26
added length of bar to bargraph config

BillKeenan:
.config/wtf/config.yml now specifies the bargraph character, and the max # of 'stars'


senorprogrammer:
🤘

PR #233, opened 2018-06-29
now will load weather api key from config if not in env

BillKeenan:
env variables are unpleasent. 

Will now load weather api key from config, if it doesn't find it in the env.

### DIFF
--- a/_sample_configs/complex_config.yml
+++ b/_sample_configs/complex_config.yml
@@ -207,6 +207,7 @@ wtf:
       cityids:
       - 3370352
       - 1283240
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Completely agree. I've had it in my mind for awhile to get rid of the ENV vars and move everything into config.

senorprogrammer:

senorprogrammer:
This is a great PR and exactly the direction I want API key/secret config to go. Thanks for this. I just made one change, which is to make the YAML key `apiKey` for case consistency and to genericize it (Jira would also then have an `apiKey` entry, for instance).

PR #234, opened 2018-07-01
gcal module improvements

kvj:
- Simplify Google OAuth client creation
- Read OAuth code via text dialog

Same approach can be re-used for gspreadsheets

### DIFF
--- a/gcal/client.go
+++ b/gcal/client.go
@@ -8,42 +8,27 @@
 package gcal
 
 import (
-	"context"
-	"encoding/json"
-	"fmt"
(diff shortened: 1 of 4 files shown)


senorprogrammer:

senorprogrammer:
I've reverted this PR because it hung for me after I entered my code at the `Authorization code:` step. I tried again, but cannot get past that step.

senorprogrammer:
Unfortunately I don't seem to be able to reopen this PR. If you could take a look at that issue and create a new PR, I'd appreciate that.

PR #236, opened 2018-07-03
Make Jira module list interactable

baustinanki:
Using the todo module as a template, this change adds interactivity
to the Jira module to allow selecting individual issues and opening
them in a web browser.

The j and k keys (or arrow keys) move up and down, and pressing
enter will open the URL for a given issue in the system's default
URL handler.

### DIFF
--- a/jira/widget.go
+++ b/jira/widget.go
@@ -3,18 +3,24 @@ package jira
 import (
 	"fmt"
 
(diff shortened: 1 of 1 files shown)


senorprogrammer:

senorprogrammer:
Beautiful.

PR #237, opened 2018-07-04
fixed bargraph test

BillKeenan:
dates were dumping 1970, added some 0's


senorprogrammer:

PR #238, opened 2018-07-06
 add todoist widget

darkSasori:
add todoist widget


senorprogrammer:

senorprogrammer:
This one is great. I can now tell Alexa to add something to a list and see it in WTF. Thanks!

PR #242, opened 2018-07-13
Add optional PR status display in Github module

baustinanki:
While spamming refresh on a pull request today to see if required
builds finished yet, it occurred to me that it'd be awesome to have
WTF tell me when a PR was ready to be merged. So, here it is! An
icon will now display next to PRs in the "My Pull Requests" section
detailing whether GitHub thinks they can be merged.

This is behind a new, opt-in config flag called "enableStatus",
due to the fact that in order to function, this feature has to hit
the GitHub API individually for each PR in order to get an updated
status check - there's a comment in the code with a link that
explains why (otherwise, `pr.GetMergeableState()` returns an empty
string). For a large number of PRs, this can slow down refreshes a
bit and _might_ even wind up rate limiting you (while testing I had
some instances of GH refusing to return me any repository info,
though it didn't actually give me an error, usually after I had
been spamming it with requests for 30 PRs in a row for a bit). So,
for that reason, use at your own risk (but it's probably fine).

I am not an emoji expert, so suggestions on the display are welcome
if you can think of anything awesome. A lot of the ones I tried
seemed to render funny and mess up spacing.


senorprogrammer:
This sounds really cool, thanks for it!

senorprogrammer:
👍

PR #248, opened 2018-07-18
fix documentation for configuring the Gerrit module

anandsudhir:
The current documentation incorrectly mentions the gerrit domain is optional. The code however requires this to be set.

This change fixes the documentation


senorprogrammer:
Can you explain to me why these changes are necessary?

senorprogrammer:
Thanks

PR #252, opened 2018-07-24
Make openFileUtil configurable

anandsudhir:
Let user configure the command/utility to use to open a file or URL from within a WTF widget.

This is useful for users that are using wtf on Linux or windows where the 'open' either doesn't exist or does something other than what open does on a mac


senorprogrammer:
Great idea.

PR #253, opened 2018-07-24
Make badges clickable and use dynamic 'all contributors' badge

anandsudhir:
- Make badges clickable
- Use dynamically generated 'all contributors' badge to avoid manual change everytime there is a new contributor


senorprogrammer:
Thanks!

PR #257, opened 2018-07-27
Fix keyboard command "o" in the todo widget

### LINKED ISSUE #256: Keyboard command "o: Open the todo file" doesn't work - atleast on Ubuntu
#### What's the problem?
In the todo mod, the keyboard command " o: Open the todo file in the operating system " doesn't seem to work. This is at least on Ubuntu. I can't see how this would work on other OS'es either.

I believe the issue is due to https://github.com/senorprogrammer/wtf/blob/master/todo/widget.go#L133.
Note that the full path of the todo.yml file isn't passed and the 'open; command cannot find the file.

Pressing 'o' when focussed on the todo widget should open the todo.yml file

anandsudhir:
Pass the full path of the todo file to openFileUtil.

Fix #256

### DIFF
--- a/todo/widget.go
+++ b/todo/widget.go
@@ -130,7 +130,8 @@ func (widget *Widget) keyboardIntercept(event *tcell.EventKey) *tcell.EventKey {
 		return nil
 	case "o":
 		// Open the file
-		wtf.OpenFile(widget.filePath)
+		confDir, _ := cfg.ConfigDir()
+		wtf.OpenFile(fmt.Sprintf("%s/%s", confDir, widget.filePath))
 		return nil
 	}
 


bertl4398:
are you sure this is going to work if you have specified an absolute path in the config.yml?
it looks to me that you always prepend the path to the configuration file also for the todo file, which is the case for the default todo.yml but not for absolute path todo.yml files ..

anandsudhir:
Hmm, I don't think you can specify absolute paths in the configuration anyway.
See https://github.com/senorprogrammer/wtf/blob/master/todo/widget.go#L209

bertl4398:
ok, thanks for clarifying .. I didn't see that in the persist method

so maybe this is also an issue but a different one, or the documentation should be more clear about that

senorprogrammer:
I consider the current file handling in the Todo module to be pretty crap:

1. There's no default value for the filename config in `widget`. There should be so that users can use Todo without having to care where the file lives.
2. The user should be able to specify an absolute path in the config if they like that over-rides the default.
3. Bug #256 is correct and `o` does not work.

senorprogrammer:
Thanks for this fix!

Issue #207, opened 2018-06-14
Todo list should scroll onscreen when moving through many items


senorprogrammer:
#### What's the problem?
When the Todo list has too many items to display onscreen, moving through the list loses the focus offscreen as it moves through items not visible. 

#### How did you expect it to behave?
The Todo list items should scroll onto the screen as the selection moves through them. The selected item should always be visible.

bertl4398:
Hi! I have fixed the scrolling by introducing tview region tags to each todo item and set the ScrollToHighlight method on the underlying textview widget .. I ll open a pull request for this issue.

senorprogrammer:
Fixed in PR #258

PR #258, opened 2018-07-27
Issue207

bertl4398:
fix for scrolling the todo list in the todo module, using tview's region tags on each todo list item and scrolling to the current highlighted list item

#207

### DIFF
--- a/todo/display.go
+++ b/todo/display.go
@@ -20,13 +20,12 @@ func (widget *Widget) display() {
 	for idx, item := range widget.list.UncheckedItems() {
 		str = str + widget.formattedItemLine(idx, item, widget.list.SelectedItem(), widget.list.LongestLine())
 		newList.Items = append(newList.Items, item)
+		offset++
 	}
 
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Scrolling works beautifully, and this solution is much better than what I had in mind, nice work. Row highlighting is a bit messed up though. Would you be able to investigate that?

screen shot 2018-07-27 at 10 03 28 am

bertl4398:
I ll have a look, sure .. actually for me it looked a little bit different with white background color on the selected item, but only if the widget had no focus (which seemed to me like a small price to pay for functioning scrolling ;) )

the documentation on the tview highlight method says that "... Text in highlighted regions will be drawn inverted, i.e. with their background and foreground colors swapped. "

https://godoc.org/github.com/rivo/tview#TextView.Highlight

but I ll see if I can improve the highlighting

bertl4398:
ok, shoud be fine now .. it was actually an easy fix, reordering the tview tags

PR #266, opened 2018-07-31
TravisCI module Pro toggle

### LINKED ISSUE #264: Support Pro version of TravisCI
The Travis CI module supports the non-Pro version of Travis CI only (travis-ci.org), it would be cool to have also the option to use the Pro version (travis-ci.com).

Add a configuration field to determine whether or not the module uses the Pro or non-Pro (default) Travis CI endpoints.

ruggi:
To fix #264, add a `pro` field to the configuration of the TravisCI module that, if `true`, uses `travis-ci.com` instead of `travis-ci.org`.


senorprogrammer:
Thanks! Nice work.

PR #267, opened 2018-07-31
Fix modals losing focus

### LINKED ISSUE #265: Modal focus is lost when typing a frame hotkey
#### What's the problem?
If a modal dialog form is open (e.g.: todo module new/edit forms), the focus on the text field is lost if a letter bound to another module hotkey (e.g.: a GitHub module) is pressed.

The focus should remain on the modal (e.g.: the form text field) until it is explicitly dismissed.

ruggi:
To fix #265, return immediately in the `FocusTracker.FocusOn` function if the current state is `AppBoard`.


senorprogrammer:
Fastest bug fix so far - thanks!

PR #270, opened 2018-08-01
Twitter module

Trinergy:
## Changes
- add twitter module for single-application dev accounts
- module documentation
- add Trinergy to contributors list
- New ENV_VARIABLE `WTF_TWITTER_BEARER_TOKEN`

## Context 📝 
This twitter module currently assumes that the dev has set up their twitter developer account
📖 https://developer.twitter.com/en/docs/basics/developer-portal/overview

And have encoded their own bearer token 
📖 https://developer.twitter.com/en/docs/basics/authentication/overview/application-only.html

## Limitations
- Bearer tokens only work for single-applications, NOT single-user developer accounts:
📖 https://developer.twitter.com/en/docs/basics/authentication/guides/single-user.html
- This current implementation only grabs tweets from a specific user


senorprogrammer:
Merged 💥

PR #273, opened 2018-08-01
Add option to hide declined calendar events

baustinanki:
My calendar view looks quite a bit cleaner with meetings I've said
"no" to taken out. This change adds a new option
`wtf.mods.gcal.showDeclined`, defaulting to `true`, which controls
whether or not the gcal module displays events where your status is
"declined".

I think as a quality of life feature, this is better off defaulting
to `false` (i.e. _don't_ show declined events by default), but when
it comes to potentially disrupting other users who've gotten used
to the existing setup, I'll leave that decision to you.


senorprogrammer:
I like that a lot and I agree, I think `showDeclined: false` should be the default. Could you make that change, and also add the config option into the docs at `_site/content/modules/gcal.md`? (I'll take care of generating the static site). Thanks for this!

PR #278, opened 2018-08-03
Make the travisci module interactive

anandsudhir:
Make the travisci module interactive
* Add ability to select and open builds
* Add commit message (oneline style) to text displayed in the widget
* Add documentation on the new keyboard shortcuts
* Cleanup some duplicate code

### DIFF
--- a/wtf/utils.go
+++ b/wtf/utils.go
@@ -113,6 +113,13 @@ func RightAlignFormat(view *tview.TextView) string {
 	return fmt.Sprintf("%%%ds", w-1)
 }
 
+func DefaultFocussedRowColor() string {
+	foreColor := Config.UString("wtf.colors.highlight.fore", "black")
+	backColor := Config.UString("wtf.colors.highlight.back", "orange")
+
+	return fmt.Sprintf("%s:%s", foreColor, backColor)
+}
+
 func RowColor(module string, idx int) string {
 	evenKey := fmt.Sprintf("wtf.mods.%s.colors.row.even", module)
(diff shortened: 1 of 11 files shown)


anandsudhir:
Failed codeclimate again? :(
Could we increase the threshold for the number of duplicate lines a bit? 

I must admit though, I am a Go noob and am trying to learn Go by working with this awesome project.

For instance, the check now fails because the below block appears twice: 
IMHO, this seems a bit too stringent.

```go
func NewWidget(app *tview.Application, pages *tview.Pages) *Widget {
    widget := Widget{
        HelpfulWidget: wtf.NewHelpfulWidget(app, pages, HelpText),
        TextWidget:    wtf.NewTextWidget("TravisCI", "travisci", true),
    }
```

senorprogrammer:
Yep, I agree. The check is too restrictive right now. I'll decrease the sensitivity.

wtf/utils.go
```diff
@@ -113,6 +113,13 @@ func RightAlignFormat(view *tview.TextView) string {
 	return fmt.Sprintf("%%%ds", w-1)
 }
 
+func DefaultFocussedRowColor() string {
```
senorprogrammer: This is a very nice improvement.

senorprogrammer:
Latest `master` has a .codeclimate.yml` that relaxes the similar code check. Give that a try.

senorprogrammer:
I've disabled CodeClimate until I have more time to dig into configuring it properly.

senorprogrammer:
Nice work, this is a great addition.

PR #279, opened 2018-08-03
Fix invalid .codeclimate.yml

anandsudhir:


anandsudhir:
Something is up with codeclimate. It doesnt seem to use .codeclimate.yml from the PR.
I guess thats why this and #278 are failing

PR #286, opened 2018-08-08
Add Gitter module

anandsudhir:
Add Gitter module

It's not perfect or the best, still a nice enough start I suppose.
We could add more (e.g. sending messages etc) to the module as we go if users want it.


senorprogrammer:
Thanks!

PR #287, opened 2018-08-09
Update ConfigDir()

Trinergy:
# Changes 🔧
- `wtf --config=path/to/your/config.yml` can now handle absolute paths (only handled relative path before)
- `cfg.ConfigDir()` now returns the directory of the custom path (used to only return default path)
- Fix typo in `config_files.go`

# Context 📖 
The purpose of these changes is so that I can use `ConfigDir()` to find the `config.yml` filepath and eventually write to it.

# Notes 📓
I created a package variable for `wtf.ConfigPath` since I can't add on top of the `wtf.Config` variable struct since it is taken from a third party package. Open to input if you prefer a different place/method of encapsulation.

### DIFF
--- a/cfg/config_files.go
+++ b/cfg/config_files.go
@@ -100,7 +100,7 @@ func CreateConfigFile() {
 
 // CreateFile creates the named file in the config directory, if it does not already exist.
 // If the file exists it does not recreate it.
-// If successful, returns the absolute path to the file
+// If successful, eturns the absolute path to the file
 // If unsuccessful, returns an error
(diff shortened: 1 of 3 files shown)


senorprogrammer:
I see what you're going for however I'm not yet sure this is the way. I think this confounds two different concepts. I'll look into this more this weekend. That said:

> wtf --config=path/to/your/config.yml can now handle 
> absolute paths (only handled relative path before)

It has always been able to handle absolute paths for defining the config file location:

> wtf --config=/Users/ccummer/Desktop/simple_config.yml

PR #292, opened 2018-08-13
Add Bearer Token resources in twitter documentation

Trinergy:
I have added the Bearer Token getter tool in a separate repo and updated the documentation.

This is an intermediate step towards automating the authentication process on initialization and writing to the `config.yml` eventually.


senorprogrammer:
Merged

PR #294, opened 2018-08-13
Hide sigils when there is only one page to display

### LINKED ISSUE #291: Sigils display for only one entry
#### What's the problem?
Some widgets use sigils to indicate the presence of more than one entry, and show the nav between them. These also show up when there's only one entry. They're not needed in that case.

Don't see the sigils when there's zero or one entry.

anandsudhir:
Sigils show up even when there are only one or no pages to display. They're not needed in these cases as there is nothing to navigate between.

Fix #291


senorprogrammer:
👍

Issue #298, opened 2018-08-17
Pretty weather prints out gibberish escape sequences in the weather symbol


anandsudhir:
#### What's the problem?
Pretty weather prints out gibberish escape sequences in the weather symbol

![image](https://user-images.githubusercontent.com/3252403/44271354-2eb1b080-a23a-11e8-9c33-6a5915642cb8.png)

#### How did you expect it to behave?
Show weather symbol without gibberish escape sequences printed

#### Operating system 
Ubuntu 18.04, Windows 10, WSL on Ubuntu 16.04

#### Terminal program & version
GNOME Terminal, cmd

#### WTF version
0.2.0-25-g02d956

bertl4398:
I fixed it .. there was a problem with the regex being parsed and replaced, made a pull request for that.

senorprogrammer:
Fixed

PR #299, opened 2018-08-17
Fix: Pretty weather prints out gibberish escape sequences in the weather symbol

bertl4398:
fix for #298


senorprogrammer:
Awesome, thank you.

PR #302, opened 2018-08-22
Add initial datadog widget

Seanstoppable:
Add a widget that will list Datadog monitors in a triggered state.


senorprogrammer:
Awesome, thanks!

Issue #307, opened 2018-08-27
Add option to configure text color and background colors of panes


ohcibi:
#### What problem does this solve?

* The project cannot be used with a light background currently as the text color cannot be changed
* Invidual background colors for panes would add the ability to increase information transmission

#### How do envision it working?
_Note that I use the word 'pane' here to refer to a single "module", or whatever it is called in wtf,  that gets displayed with a border. If unclear what I mean please ask instead of assume as I don't already know the wording in this project and I very likely have used the wrong term._

* The text color could be an option similar to the current `color.background` e.g. `color.foreground` (if this is already available, consider this issue as a documentation improvement request)
* Styling panes is a bit more of a complex problem. Possible options, I could think of are
  * Text/title matching of text inside the pane or of individual values of the pane
    * would allow something like e.g. using `red` background for the bamboo pane if the latest job has failed to build
  * Adding something like a `class` (as in "css class") to the meta information of a pane and match based on that
  * positional colorization (e.g. top row of panes is blue, center row is yellow, bottom row is red)

jaimejim:
I would personally some generic mechanism for changing the colors of each item within a module. There are some global variables that can be used in the `config.yaml`, so far I know of:

```
text, subheading
```
And the use of `::b` to make the text bold.

Are these things documented somewhere? I at least could not find them on https://wtfutil.com/configuration/global_settings/

PR #320, opened 2018-09-27
Add option to set background and text color on a module basis

nicholas-eden:
Resolves the feature request in issue #307 .

Each mod can use a `colors` namespace to set the `background` and `text` properties, as which will default to the global value if not set.

```
wtf:
  colors:
    text: "gray"
    border:
      focusable: darkslateblue
      focused: orange
      normal: gray
  grid:
    columns: [40, 40]
    rows: [13, 13, 4]
  refreshInterval: 1
  mods:
    clocks:
      colors:
        background: "darkred"
        rows:
          even: "lightblue"
          odd: "white"
      enabled: true
      ...
    security:
      colors:
        text: "yellow"
      enabled: true
      ...

```

selection_006
]


senorprogrammer:
Thanks!

PR #323, opened 2018-10-02
Clock format

danrabinowitz:
I prefer 12-hour am/pm times instead of 24-hour time. This PR adds two new optional attributes for the clock module, to support custom formats for date and time.


senorprogrammer:
Thanks @danrabinowitz, this is an excellent improvement.

PR #325, opened 2018-10-04
Display all day events in calendar

nicholas-eden:
All day events were being excluded from the calendar.  It looks like they were excluded due to time zone issues, parsing the date would result in a UTC time which would display under the previous date due to the time zone difference.

selection_009

All day events will display without a time.


senorprogrammer:
Thanks!

PR #327, opened 2018-10-05
Suspend app to make authenticating for google calendar easier

dvdmssmnn:
The initial setup for google calendar is quite complicated.
With this pull request the app gets suspended when the tokens are not available and the link for authenticating is printed.

Unfortunately, there is still a small (known) bug in `tview` (https://github.com/gdamore/tcell/issues/194) that looses a keyboard event after suspending the app, which requires users to hit enter before inserting the approval code from google.

Also, the tokens are now stored in `.config/wtf/` instead of `~./credentials`.


senorprogrammer:
Fantastic improvement, thanks for this!

PR #329, opened 2018-10-11
add mercurial module

### LINKED ISSUE #328: Mercurial Support
I would like to have an integration for mercurial as there is one for git. If it is ok i will try to create this as a new module with the same functionality as Git.

mweb:
I added a module for mercurial. This is copy of the git module with the required adaption for mercurial. Let me now if I need to change anything. I will provide the documentation within the wtfdocs repository.
closes #328


senorprogrammer:
That's better 👍

senorprogrammer:
Thanks for adding this!

Issue #305, opened 2018-08-22
Spotify controller module


StormFireFox1:
#### What problem does this solve?
I have recently downloaded this program and have used it consistently on all my devices. One thing I do find important in my developer workflow, however, is music. It would be cool if WTF had a music player (possibly linked with Spotify) module that could be added on the grid.

#### How do envision it working?
It would be a module that would contain the media player's current song info, etc (no need for album image) and would have a few keys (like p, left and right arrow) to play/pause, change song, etc. The integration with Spotify would probably be a lot easier than it seems, as well. (considering there are libraries to interface with Spotify in Go)

senorprogrammer:
That would be awesome. I hope someone builds this.

anandsudhir:
Can this be closed now?  #317 has been merged

senorprogrammer:
Good idea. I'll do that as soon as I add some documentation for the Spotify module.

senorprogrammer:
Added by @sticreations sticreations.

StormFireFox1:
Thank you very much for this! Whenever I get time, I'll try to contribute and add Windows support for this module.

senorprogrammer:
That would be great. Looking forward to rolling that in @StormFireFox1

sticreations:
I actually wanted to do that, but it seems that OSX and Windows don’t have something like linuxes dbus. So you would have to use the Spotify API. 
My goal was to have as little configuration as possible, but I think using the api would be way more easier to generalize than finding a hacky way to control it via something else.

StormFireFox1:
I found a Web API for Spotify written in Go:

https://github.com/zmb3/spotify

This should technically work, but for a normal user, it takes time to configure that some might not put in.
If I find a way to make the current version of the widget work with Windows as well, I'll do that. Otherwise, the package linked seems reasonable enough.

voldemortensen:
I submitted a PR to @sticreations package that adds OSX support via apple script: https://github.com/sticreations/spotigopher/pull/3

senorprogrammer:
@StormFireFox1 That looks interesting indeed. I would support the adoption of either a) a single Spotify module that can be configured to work with either the web API or the local client, or b) two Spotify modules, one for the web API and one for the local client.

PR #332, opened 2018-10-13
Add Spotify Web Widget

StormFireFox1:
---
name: Add Spotify Web API Widget
about: Adding Spotify integration
---

This adds a version of the Spotify widget that uses the Spotify Web API.
It is as close to a replica as the original Spotify widget, that uses Linux d-buses.

#### Is this related to an existing Issue? If so, which one?

This solves issue #305 again.


senorprogrammer:
Not ignoring this one, just out of town. I'll try to get it in today.

senorprogrammer:
Thanks for the PR!

PR #333, opened 2018-10-15
Added firewall status on linux

TheRedSpy15:
This does require sudo, however it will just return "NA" if otherwise


senorprogrammer:
Thanks for this!

Issue #313, opened 2018-09-06
Config validator


senorprogrammer:
#### What problem does this solve?
It's easy to introduce formatting errors into the config file when editing manually and when the app stops working, it's not always obvious where the error is. 

WTF should be able to parse its own config and report back any errors.

#### How do envision it working?

On launch and on each config reload, parse it. If any errors are present, display that information in a window.

senorprogrammer:
Currently does this for Position elements on all modules.

PR #334, opened 2018-10-15
add static hot key assignments for text widgets

### LINKED ISSUE #330: Allow static hot key assignments
Currently hot key numbers are dynamically assigned to widgets, which means the more interactive widgets you have, the less likely they are to have the same number between executions (or config reloads). This makes it impossible to develop muscle memory for widgets.

Allow the hot key number to be assigned in the config file for the widget. If assigned, it is removed from the pool of dynamically-assigned numbers available for use.

mweb:
This commit allows configuring static hot key assignments. Just add the key focusChar to a focusable widget and the predefined focusChar will be set.
This closes #330 

There is one open problem if two widgets are configured with the same focusChar then only one can be accessed via the hot keys. But before I resolve this I would like to get a first feedback to the change.

For the duplicate focusChar I see the following solutions:

1. Leave as it is, it is visible to the user that there are multiple widgets with the same focusChar
2. If there are multiple widgets with the same focusChar, use the next free hot key
3. If there are multiple widgets with the same focusChar, show an error (but this is something for #313)

### DIFF
--- a/wtf/focus_tracker.go
+++ b/wtf/focus_tracker.go
@@ -29,9 +29,25 @@ func (tracker *FocusTracker) AssignHotKeys() {
 		return
 	}
 
+	usedKeys := make(map[string]bool)
+	focusables := tracker.focusables()
 	i := 1
 
-	for _, focusable := range tracker.focusables() {
+	for _, focusable := range focusables {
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Thanks!

PR #336, opened 2018-10-28
Bump Go versions and use '.x' to always get latest minor versions

HaraldNordgren:


.travis.yml
```diff
@@ -1,7 +1,7 @@
 language: go
 go:
-  - "1.10"
-  - "1.11"
+  - "1.10.x"
+  - "1.11.x"
```
HaraldNordgren: Just realized something that I forgot in the last PR!

This bumps the 1.10 patch version from `go1.10 linux/amd64` to `go1.10.4 linux/amd64`. 1.10 is implicitly 1.10.0 -- try `eval "$(gimme "1.10.0")"` to see that there is no 1.10.0 version. Same for 1.11.

PR #340, opened 2018-11-13
Add shortcut to open Hacker News comments

kaskavalci:
It's nice to read HN articles from wtf but it's not possible to see discussions. `c` keyword will open the relevant comment thread of the article.


anandsudhir:
Thanks :)

PR #361, opened 2019-01-20
Go generate

retgits:
_as mentioned in #310_

## What problem does this solve?
Right now anyone creating a new module has to start from scratch, probably copy/pasting/chopping the guts from an existing module. This is tedious, error-prone, and not a great use of people's time.

## How does it work?
The package `generator` contains generates for empty widgets (stubs). Each generator is named after the type of widget it generate, so `textwidget.go` will generate the skeleton for a new TextWidget using the `textwidget.tpl` template. Since you can't dynamically add any arguments to `go generate`, the TextWidget generator needs one environment variable, called _**WTF_WIDGET_NAME**_, which will be the name of the TextWidget it generates. If the variable hasn't been set, the generator will use `NewTextWidget`. As an example, on macOS the command can be run as `WTF_WIDGET_NAME=MySuperAwesomeWidget go generate -run=text`, which will generate a new module based on the TextWidget generator in the folder `mysuperawesomewidget`

### DIFF
--- a/generator/textwidget.go
+++ b/generator/textwidget.go
@@ -5,7 +5,7 @@
 // using the textwidget.tpl template.
 // The TextWidget generator needs one environment variable, called WTF_WIDGET_NAME, which will
 // be the name of the TextWidget it generates. If the variable hasn't been set, the generator
-// will use "NewWiNewTextWidgetdget". On Linux and macOS the command can be run as
+// will use "NewTextWidget". On Linux and macOS the command can be run as
 // 'WTF_WIDGET_NAME=MyNewWidget go generate text'.
 package main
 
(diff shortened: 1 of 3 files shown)


senorprogrammer:
This is awesome, and I would love to merge this in. Before that though, I'm getting the following error when I follow the directions above:
```
~/go/src/github.com/wtfutil/wtf retgits-go-generate 8s
❯ WTF_WIDGET_NAME=MySuperAwesomeWidget go generate text
can't load package: package text: no Go files in /usr/local/Cellar/go/1.11.1/libexec/src/text
```
Suggestions?

retgits:
Totally my bad, sorry. The command should be `WTF_WIDGET_NAME=MySuperAwesomeWidget go generate -run=text`. I'll update the text so it reflects the correct command.

senorprogrammer:
Thanks!

retgits:
All updates should be in place now 😄

PR #365, opened 2019-01-30
Changed nmcli cmd to check on AP connect to using '*' in the in-use f…

jdenoy:
Changed nmcli cmd to check on AP connect to using '*' in the in-use field instead of 'yes/no' from the active field. This is needed for non english language set in operating system.

This is needed because yes/no will not be the same text depending on the language configured in the operating system. Using the "in-use" field that is either set to "*" ot "" will be easier for the "wtf.FindMatch" function to match.


senorprogrammer:
Very nice, thanks for improving this.

Issue #364, opened 2019-01-28
Opsgenie not working


GaboFDC:
#### What's the problem?
Opsgenie api changed, the on-call endpoint was deprecated https://docs.opsgenie.com/docs/who-is-on-call-api#section-list-on-calls

```bash
curl -X GET 'https://api.opsgenie.com/v2/schedules/on-calls' --header 'Authorization: GenieKey xx-xx-xx-xx-xx'

{"message":"On-calls endpoint is not supported anymore.","took":0.007,"requestId":"xxx-5213-438c-91d6-xxx}
```

#### How did you expect it to behave?
Have to change the implementation to use new api

#### Operating system - Terminal program & version - WTF version
not relevant

senorprogrammer:
Unfortunately I don't use OpsGenie anymore and have no API access so someone else will have to solve this one.

senorprogrammer:
Thanks @GaboFDC, much appreciated

PR #366, opened 2019-01-31
Use new api format

GaboFDC:
I'm super new to go (and even more to pointer), I tried to follow the same patters that the code had, but also maybe i did some strange things.
If something can be improved I'm open to suggestions.

Fixes https://github.com/wtfutil/wtf/issues/364
Now you have to specify `scheduleIdentifierType` and `schedule` in the config.

If you wanna see more schedules you can pass a list: `schedule: ["scheduleA", "scheduleB"]`

Also i'm not sure if open a PR updating the documentation right now, or wait til the PR is ready to merge

### DIFF
--- a/opsgenie/client.go
+++ b/opsgenie/client.go
@@ -10,10 +10,10 @@ import (
 )
 
 type OnCallResponse struct {
-	OnCallData []OnCallData `json:"data"`
-	Message    string       `json:"message"`
-	RequestID  string       `json:"requestId"`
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Sweet, much appreciated. I will try and look at this this weekend and get it merged in but I just had some surgery, so it might take me a few extra days to get to it; if it does, rest assured thatches is not being ignored.

senorprogrammer:
Thanks for these improvements!

PR #387, opened 2019-02-15
Add default unknown widget

Seanstoppable:
Playing off of https://github.com/wtfutil/wtf/pull/383
Rather than silently failing, let people know if they are doing something a bit wrong


anandsudhir:
Thanks. I agree, this is a sweet spot between failing fast and absolute silence :)

anandsudhir:
Oops. Closed without merging. Sorry about that

PR #388, opened 2019-02-15
PoC implementation for allowing dupes

Seanstoppable:


senorprogrammer:
What's the plan for this PR? Should it stick around?

Seanstoppable:
I plan to revisit now that we've refactored config significantly. I think it should make some of this slightly easier.

senorprogrammer:
Duplicate instances of modules is one of the main motivators for that change. To make dupes possible what do you think of this change to the configuration?

Right now modules are defined like:

```
mod:
    clocks:
        enabled: true
        refreshInterval: 300
        ...
       etc.
```
The module name is hard-coded as the key in the map. Instead, move the module name into the configuration itself:
```
mod:
    firstClocks:
        enabled: true
        module: clocks
        refreshInterval: 300
        ...
    secondClocks:
        enabled: true
        module: clocks
        refreshInterval: 150
```
To preserve backwards-compatibility, if the `module` key was missing, it would use the parent key as the module name.

Seanstoppable:
I guess it has to be, since the config map will flatten keys. I'll have another PoC this weekend

PR #394, opened 2019-02-26
Decouple newrelic client and widget

Seanstoppable:
Have widget manage config
Have client handle communication


senorprogrammer:
Thank you

PR #395, opened 2019-02-26
Decouple CircleCI client from widget

Seanstoppable:
Have widget manage config, have client just manage communication


senorprogrammer:
Very nice

Issue #358, opened 2019-01-19
Jenkins Green Balls


skymeyer:
Successful builds in the Jenkins module are using the blue color. Would like to add the option to use green instead of blue to mimic https://wiki.jenkins.io/display/JENKINS/Green+Balls.

**Proposal**
* Add a boolean greenBalls config parameter to use green instead of blue in reference to the green balls plugin
* Or add config parameters to be able to customize each colors

Thoughts ?

rudolphjacksonm:
Hi folks--I'm learning Go and this seems like a good issue to take a stab at. @skymeyer having a boolean config parameter seems like the easiest place to start. Are you ok with me taking a look at this?

rudolphjacksonm:
Also opened a PR for the wtfutildocs, see https://github.com/wtfutil/wtfdocs/pull/21

senorprogrammer:
Thanks for tackling this one!

senorprogrammer:
Closed with https://github.com/wtfutil/wtf/pull/403

PR #403, opened 2019-04-15
Jenkins Green Balls Plugin Support - #358 

rudolphjacksonm:
Support for #358 - Adds successBallColor config value which supports custom Jenkins job colour to indicate success instead of default blue.

### DIFF
--- a/modules/jenkins/widget.go
+++ b/modules/jenkins/widget.go
@@ -120,7 +120,12 @@ func (widget *Widget) rowColor(idx int) string {
 func (widget *Widget) jobColor(job *Job) string {
 	switch job.Color {
 	case "blue":
-		return "blue"
+		// Override color if greenBalls boolean param provided in config
+		if wtf.Config.UString("wtf.mods.jenkins.greenBalls") == "true" {
+			return "green"
+		} else {
+			return "blue"
+		}
 	case "red":
 		return "red"
 	default:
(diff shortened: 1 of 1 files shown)


modules/jenkins/widget.go
```diff
@@ -120,7 +120,12 @@ func (widget *Widget) rowColor(idx int) string {
 func (widget *Widget) jobColor(job *Job) string {
 	switch job.Color {
 	case "blue":
-		return "blue"
+		// Override color if greenBalls boolean param provided in config
+		if wtf.Config.UString("wtf.mods.jenkins.greenBalls") == "true" {
+			return "green"
+		} else {
+			return "blue"
+		}
```
senorprogrammer: What do you think about extending this out to support whatever coloured balls the user wants, defaulting to blue?

Something like this in the config YAML:

```
    wtf.mods.jenkins.ballColor: "green"
```

and in code:

```
    wtf.Config.UString("wtf.mods.jenkins.successBallColor", "blue")
````

Comment (senorprogrammer) [reply modules/jenkins/widget.go]: I'm also assuming those balls indicate success. I don't use Jenkins so if they're indicating something else, the name would reflect that.

Comment (rudolphjacksonm) [reply modules/jenkins/widget.go]: Yep you're exactly right, the balls indicate the status of the job. By default Jenkins uses blue (don't ask me why). I've knocked this up but if this isn't best practice or there's a better way of doing this you might have to hold my hand a little here!
```
case "blue":
		// Override color if greenBalls boolean param provided in config
		ballColor := wtf.Config.UString("wtf.mods.jenkins.successBallColor", "blue")
		if ballColor != "blue" {
			return ballColor
		} else {
			return "blue"
		}
```

senorprogrammer: Looks good to me

senorprogrammer:
Thanks!

Issue #359, opened 2019-01-19
Jenkins regex job filter


skymeyer:
I'm looking into adding a regex job filter to control which jobs are being listed.

**Proposal**
* Add optional regex filter field

senorprogrammer:
Added with https://github.com/wtfutil/wtf/pull/405

PR #405, opened 2019-04-17
Allow users to filter jenkins jobs by regex

rudolphjacksonm:
Addresses issue #359. Allows users to filter the Jenkins jobs shown in the widget by providing a regular expression in the config file. The regex should be specified in the config like so:
```
jobNameRegex: ^[a-z]+.$
```
Another example:
```
jobNameRegex^[a-z]+\[[0-9]+\]$
```

### DIFF
--- a/modules/jenkins/widget.go
+++ b/modules/jenkins/widget.go
@@ -2,6 +2,7 @@ package jenkins
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Sweet. I'll try to get this in today. Thanks!

PR #410, opened 2019-04-23
20190422 signal redraw on channel

senorprogrammer:
This swaps out the hard-coded `app.Draw()` call in `TextWidget` for use of `refreshChan`, widgets signalling on a channel that they want the app to redraw.

This simplifies some modules and complicates others. It remains to be seen if this is a good idea.

### DIFF
--- a/cfg/common_settings.go
+++ b/cfg/common_settings.go
@@ -55,9 +55,10 @@ type Common struct {
 	Sigils
 
 	Enabled         bool
-	FocusChar       int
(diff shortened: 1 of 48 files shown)


senorprogrammer:
First tests of this branch seem to indicate that the channels version requires far more CPU than non-channels, enough so as a completely unscientific test, my laptop fans spin up on this branch within seconds, and on master do not at all.

Something is wrong.

PR #419, opened 2019-04-27
Decouple modules from global config

Seanstoppable:
Rather than referencing wtc.Config, instead pass the global config
Also, look up config for the module early and pass that in sooner, to deal with fewer long paths and get rid of the
ConfigKey variable


senorprogrammer:
Excellent improvement!

PR #420, opened 2019-04-28
Update title handling so that config based title can be used

Seanstoppable:
There is already default handling of title. Remove instances where people are unknowingly overriding it
For instances where we want to set special things, make sure to use CommonSettings.Title, so people can still override


modules/nbascore/widget.go
```diff
@@ -50,7 +50,6 @@ func NewWidget(app *tview.Application, pages *tview.Pages, settings *Settings) *
 func (widget *Widget) Refresh() {
 	widget.app.QueueUpdateDraw(func() {
 		widget.nbascore()
-		widget.View.SetTitle(widget.ContextualTitle(widget.Name()))
```
senorprogrammer: This change breaks the title if it has been assigned a numeric hot key (all widgets need to render the title through `ContextualTitle` in case they've got a `FocusChar`).

This only applies to widgets that are `Focusable`, though for future-proofing all widget titles should probably just go through `ContextualTitle`.

modules/opsgenie/widget.go
```diff
@@ -44,7 +44,6 @@ func (widget *Widget) Refresh() {
 	}
 
 	widget.app.QueueUpdateDraw(func() {
-		widget.View.SetTitle(widget.ContextualTitle(widget.Name()))
```
senorprogrammer: This one will probably also break hot key titles as well, etc.

senorprogrammer:
This is a nice improvement towards consistency, just a few spots that it breaks existing UI.

modules/nbascore/widget.go
Seanstoppable: I thought the general case did this anyway, but I will explicitly test to understand the difference. Thanks for pointing this out!

senorprogrammer: Yeah that makes much more sense. Thanks for looking into this.

senorprogrammer: Note that any calls to `SetTitle` must be wrapped in either `QueueUpdate()` or `QueueUpdateDraw` https://github.com/rivo/tview/wiki/Concurrency

senorprogrammer:
The title issue happens because of order of operations:

* main.go calls for widget creation
* widgets are created but have no FocusChar assigned to them
* TextWidget sets the default title but does so without the FocusChar because of the above
* main.go calls for NewFocusTracker
* FocusTracker rolls through the widgets and sets the FocusChar for all focusable ones
* TextWidget's title is never re-called to render with the FocusChar

senorprogrammer:
I've re-applied the titles in this commit https://github.com/wtfutil/wtf/commit/378cce87269c700be0267591dc89b202b38e0d03 and merged this branch into master

PR #424, opened 2019-05-03
Fix up Unknown widget to work again

Seanstoppable:
Store module Type in common settings to enable this
Simplify the function params to be more inline with existing ones, vs being unique


senorprogrammer:
Nice!

Issue #429, opened 2019-05-07
Race Conditions in widget Rendering


Seanstoppable:
#### What's the problem?

Running with `-race` occasionally triggers a race, even if I've disabled config watching. In regular operation, I suspect this causes wtf to silently stop updating

race output
```
==================
WARNING: DATA RACE
Read at 0x00c0003ee251 by main goroutine:
  github.com/rivo/tview.(*TextView).Draw.func1()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/textview.go:926 +0xecc
  github.com/rivo/tview.iterateString.func1()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/util.go:515 +0x1fa
  github.com/rivo/tview.iterateString()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/util.go:527 +0x2d6
  github.com/rivo/tview.(*TextView).Draw()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/textview.go:874 +0x783
  github.com/rivo/tview.(*Grid).Draw()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/grid.go:583 +0x2e22
  github.com/rivo/tview.(*Pages).Draw()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/pages.go:246 +0x190
  github.com/rivo/tview.(*Application).draw()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/application.go:363 +0x164
  github.com/rivo/tview.(*Application).QueueUpdateDraw.func1()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/application.go:494 +0x54
  github.com/rivo/tview.(*Application).Run()
      /Users/ssmith/go/pkg/mod/github.com/rivo/tview@v0.0.0-20181226202439-36893a669792/application.go:253 +0x603
  main.main()
      /Users/ssmith/code/personal/wtf/main.go:151 +0xa1b

Previous write at 0x00c0003ee251 by goroutine 14:
[...]

PR #430, opened 2019-05-08
Add a global Redraw method for TextWidget

Seanstoppable:
Partially addresses #429, by centralizing widget drawing

### DIFF
--- a/modules/circleci/widget.go
+++ b/modules/circleci/widget.go
@@ -11,7 +11,6 @@ type Widget struct {
 	wtf.TextWidget
 	*Client
 
-	app      *tview.Application
 	settings *Settings
 }
 
@@ -20,7 +19,6 @@ func NewWidget(app *tview.Application, settings *Settings) *Widget {
 		TextWidget: wtf.NewTextWidget(app, settings.common, false),
 		Client:     NewClient(settings.apiKey),
 
-		app:      app,
 		settings: settings,
 	}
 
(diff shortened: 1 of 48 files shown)


modules/circleci/widget.go
```diff
@@ -11,7 +11,6 @@ type Widget struct {
 	wtf.TextWidget
 	*Client
 
-	app      *tview.Application
```
Seanstoppable: Trying to eliminate this being stored here as much as possible

modules/gcal/widget.go
```diff
@@ -61,27 +53,5 @@ func (widget *Widget) fetchAndDisplayEvents() {
 		widget.calEvents = calEvents
 	}
 
-	widget.app.QueueUpdateDraw(func() {
-		widget.display()
-	})
-}
-
-func updateLoop(widget *Widget) {
```
Seanstoppable: This seemed unnecessary, as the refresh of the widget itself should handle this, no?

modules/gitter/widget.go
```diff
@@ -102,11 +94,10 @@ func (widget *Widget) display() {
...
 
-	widget.View.SetWrap(true)
-	widget.View.Clear()
-	widget.View.SetTitle(widget.ContextualTitle(fmt.Sprintf("%s - %s", widget.CommonSettings.Title, widget.settings.roomURI)))
-	widget.View.SetText(widget.contentFrom(widget.messages))
-	widget.View.Highlight(strconv.Itoa(widget.selected)).ScrollToHighlight()
+	title := fmt.Sprintf("%s - %s", widget.CommonSettings.Title, widget.settings.roomURI)
+
+	widget.Redraw(title, widget.contentFrom(widget.messages), true)
+	//widget.View.Highlight(strconv.Itoa(widget.selected)).ScrollToHighlight()
```
Seanstoppable: I don't have any widgets with highlight, so not 100% sure what is going on here and/or if this is a potential race condition that we should also try to abstract away

modules/gcal/widget.go
senorprogrammer: Yeah, I'm not sure why that's there. It should be removed.

senorprogrammer:
I've found a few small issues with this branch that need to be fixed first, otherwise its looking good:

* some widgets with nav keys assigned to them display the key number twice in the title
* the "new item" modal for Todo takes a long time to show up because the explicit `Draw()` that forced an update has been removed

I should be able to get to those today.

Seanstoppable:
No problem. After this, I am going to try refactoring scrollable widgets to remove boilerplate from them too

senorprogrammer:
All the changes from this PR have been merged https://github.com/wtfutil/wtf/commit/018d2af3ae4c17a93e6a0da4881263d6d993f577

PR #432, opened 2019-05-08
fix(gcal): remove possible extraneous variation codepoint

gnanderson:
My layout was breaking:

gcal_layout

I found an extraneous codepoint after the check mark glyph

    ' ︎' U+FE0E Dec:65038 VARIATION SELECTOR-15 &#xFE0E; /\%ufe0e "\ufe0e"

This PR removes it.


senorprogrammer:
Nice catch, I don't see that on mine at all. Thanks!

Issue #431, opened 2019-05-08
No line highlighting in JIRA/possibly others


Seanstoppable:
#### What's the problem?

When scrolling through the JIRA module, I don't see lines highlighted, and so it isn't clear which line I might be interacting with to open in browser

#### How did you expect it to behave?

Highlighting, so I know what line I am on.

#### Operating system 

OSX

#### Terminal program & version

iTerm2 and Terminal

#### WTF version

master

PR #433, opened 2019-05-09
Move closing region tag to the end of the line

Seanstoppable:
Highlight works by highlighting the entire region that exists
Since current region starts and immediately ends, there is no highlighting
Addresses #431


senorprogrammer:
Thanks!

PR #434, opened 2019-05-09
feat(jira): select first result if exists

gnanderson:
select the first jira result: a) if it exists, b) if the widget has just been instantiated (selected < 0)

Inspired by https://github.com/wtfutil/wtf/pull/433 :)

### DIFF
--- a/modules/jira/widget.go
+++ b/modules/jira/widget.go
@@ -75,6 +75,9 @@ func (widget *Widget) Refresh() {
 		})
 	} else {
 		widget.result = searchResult
+		if len(widget.result.Issues) > 0 && widget.selected < 0 {
+			widget.selected = 0
+		}
 	}
 
 	widget.app.QueueUpdateDraw(func() {


modules/jira/widget.go
```diff
@@ -75,6 +75,9 @@ func (widget *Widget) Refresh() {
 		})
 	} else {
 		widget.result = searchResult
+		if len(widget.result.Issues) > 0 && widget.selected < 0 {
+			widget.selected = 0
```
Seanstoppable: A few things, having recently dived into the selection code:
1. This means that every time the widget refreshes, it will try to highlight the first row (highlighting is broken right now, but should get fixed soonish)
1. Rather than starting with things not selected (default case right now, with selected defaulting to -1), we instead select the first row
In both the above cases, this is perhaps not a great user experience. By default, I think I prefer a stock list without highlighting (which can be cleared), vs having a row always highlighted
1. This also means that when things are refreshed, it will reset where the positioning was to the first item again. I can see this being frustrating if you are scrolling through and hit a refresh.

senorprogrammer: @gnanderson Your thoughts on the comments above?

senorprogrammer:
@gnanderson If there's no feedback from you by July 9 on the comments in this PR I'll assume it's unmaintained and close it out.

senorprogrammer:
Closing this PR out as unmaintained.

PR #441, opened 2019-05-10
Split up logger and widget

Seanstoppable:
This allows us to use the logger from the wtf directory
For example when trying to debug sort ordering in focus_tracker


senorprogrammer:
Looks like `app` should now be removed from DataDog:

```
# github.com/wtfutil/wtf/modules/datadog
modules/datadog/widget.go:39:3: unknown field 'app' in struct literal of type Widget
The command "go get ./... && go get github.com/go-test/deep && go test -v github.com/wtfutil/wtf/wtf_tests/..." exited with 2.
```

PR #442, opened 2019-05-11
Fix datadog rendering

Seanstoppable:
This needs app until the scrollable refactor is done


senorprogrammer:
Scrollable refactor just went in now.

PR #443, opened 2019-05-11
Have keyboard widget manage its own help

Seanstoppable:
Define help with keys
This means that keys and help are automatically in sync
This means that you can't define keys, but forget help
This unfortunately also means that formatting may not be quite as good


Seanstoppable:
Example of how it looks
Screen Shot 2019-05-11 at 12 14 22 PM

flags/flags.go
```diff
@@ -6,13 +6,11 @@ import (
 	"path/filepath"
 
 	goFlags "github.com/jessevdk/go-flags"
-	"github.com/wtfutil/wtf/help"
 	"github.com/wtfutil/wtf/utils"
 )
 
 type Flags struct {
 	Config  string `short:"c" long:"config" optional:"yes" description:"Path to config file"`
-	Module  string `short:"m" long:"module" optional:"yes" description:"Display info about a specific module, i.e.: 'wtf -m=todo'"`
```
Seanstoppable: I ripped this out as it doesn't seem as useful/it is not getting updated with new modules.

senorprogrammer: This allows someone to view the help text of a module from the command line without having to create a config for the module and run the app, eg: `wtf -m=todo` will display the help text for the Todo module. As long as new modules have help text in them, this works.

I don't want to lose this. Help text as it is now is the most basic implementation possible, it could be much more than just "this key does that" - think `man` pages for WTF modules, including configuration instructions, API links, etc. I just haven't spent any time on that side of things yet.

senorprogrammer:
I like this direction however the display will have to be improved. That's not a great look. `HelpItem` could have something like a `type` field ('key' or 'char'), and split the groupings on that, for instance.

Seanstoppable:
That is fair.

I have some ideas for both formatting and making the manpages portion more
of a reality.

senorprogrammer:
Thanks!

Issue #408, opened 2019-04-22
Sample configs need to be updated


senorprogrammer:
#### What problem does this solve?

The same configs are wildly out of date with a basic configuration.

retgits:
I'd love to help out on this one, though when I run the configs they seem to work very well... what kind of config would you like to see?

senorprogrammer:
For the modules included in the samples they should have a complete set of all the available options for that module (the same things needs to happen in the docs - a lot of modules have config settings that aren't yet reflected in the docs).

And as of a few minutes ago and the merge of PR #418, a config that shows multiple instances of the same module(s) needs to be added.

senorprogrammer:
Any help on this is greatly appreciated.

Seanstoppable:
So, I was thinking that we could generate these, since every module now has its own settings class.

Following along with [this](https://sosedoff.com/2016/07/16/golang-struct-tags.html), we could define struct tags of `description`, `example` and `optional` to generate a chunk of the config.

This actually will also enable 'setup walkthrough', so you could do something like `wtf generate jira`, and it will walk you through the setup, reducing manual errors.

More advanced examples (i.e. https://github.com/wtfutil/wtfdocs/blob/master/content/modules/jira.md) will still be needing to be done manually, but that is where manual work is worth it.

senorprogrammer:
"This actually will also enable 'setup walkthrough', so you could do something like wtf generate jira, and it will walk you through the setup, reducing manual errors."

This is an awesome idea.

retgits:
I'd love to take a stab at adding that feature 😄 sign me up for it

retgits:
@senorprogrammer I created a new version of `simple_config` [here](https://github.com/retgits/wtf/blob/master/_sample_configs/simple_config.yml), is that in line with your thinking for the updated configs? It contains two clocks and the module config is updated with the info I could find in the docs and settings

senorprogrammer:
Yeah, that looks really good. I'd also like to increase the size of the rows/columns to give people an idea on how to set it up for a flexible screen. 2x3 is small enough that people are still confused by it (admittedly the grid aspect is not the most intuitive thing).

For instance, my config is:
```
wtf:
    grid:
        columns: [35, 35, 35, 35, 35, 35]
        rows: [10, 10, 10, 10, 10, 3, 4]
```
[...]

PR #446, opened 2019-05-11
Update sample configs

retgits:
Updated the sample configs and added screenshots so users can see what the end results would be when using the particular configuration.

Fixes: #408


senorprogrammer:
Fantastic, thanks!

PR #447, opened 2019-05-12
Allow PagerDuty to restrict on scheduleID

senorprogrammer:
At DO we have so many schedules that unfiltered access to them causes PagerDuty to return 429 instead of data.

This PR permits filtering on `scheduleID` to restrict data to specific schedules.

Configuration example:
```
    pagerduty:
      apiKey: "xxLxxxxxxx7Zxx-Dxxx"
      enabled: true
      position:
        top: 2
        left: 3
        height: 2
        width: 1
      refreshInterval: 30
      scheduleIDs:
      - "P8I54CY"
      - "P684E4S"
      - "PWW6L2J"
      showIncidents: false
      showSchedules: true
```


modules/pagerduty/client.go
```diff
@@ -7,14 +7,17 @@ import (
 )
 
 // GetOnCalls returns a list of people currently on call
-func GetOnCalls(apiKey string) ([]pagerduty.OnCall, error) {
+func GetOnCalls(apiKey string, scheduleIDs []string) ([]pagerduty.OnCall, error) {
```
Seanstoppable: So, it probably makes sense to support the rest of the query options at some point
Do we just take a ListOnCallOptions as a param? Or just worry about that later?

modules/pagerduty/client.go
```diff
@@ -7,14 +7,17 @@ import (
...
 
 	var results []pagerduty.OnCall
-
 	var queryOpts pagerduty.ListOnCallOptions
-	queryOpts.Since = time.Now().Format("2006-01-02T15:04:05Z07:00")
-	queryOpts.Until = time.Now().Format("2006-01-02T15:04:05Z07:00")
+
+	queryOpts.ScheduleIDs = scheduleIDs
+
+	timeFmt := "2006-01-02T15:04:05Z07:00"
```
Seanstoppable: Good refactor, should probably be a const now that I see it.

modules/pagerduty/settings.go
```diff
@@ -14,6 +14,7 @@ type Settings struct {
 
 	apiKey           string
 	escalationFilter []interface{}
+	scheduleIDs      []interface{}
```
Seanstoppable: Tying into supporting other filters, should these all be top level, or should we have a `filtering` config, which these are under?

PR #449, opened 2019-05-13
Add a hierarcy of elements

Seanstoppable:
This ensures that all textwidget elements have help as well as some default functionality
This ensures that all scrollable elements have help items for the controls of that widget


maker/widget_maker.go
```diff
@@ -66,34 +66,34 @@ func MakeWidget(
 	switch widgetType {
 	case "bamboohr":
 		settings := bamboohr.NewSettingsFromYAML(widgetName, moduleConfig, globalConfig)
-		widget = bamboohr.NewWidget(app, settings)
```
Seanstoppable: Added benefit, widgets are converging on method signatures.

wtf/keyboard_widget.go
```diff
@@ -28,8 +29,8 @@ type KeyboardWidget struct {
 }
 
 // NewKeyboardWidget creates and returns a new instance of KeyboardWidget
-func NewKeyboardWidget(app *tview.Application, pages *tview.Pages, settings *cfg.Common) KeyboardWidget {
-	return KeyboardWidget{
+func NewKeyboardWidget(app *tview.Application, pages *tview.Pages, settings *cfg.Common) *KeyboardWidget {
```
Seanstoppable: I got weird functionality before turning this into a pointer.
I would only get part of the help screen
Some of the commands would try to work, but fail on null pointers
But, to be honest, I don't understand WHY. Does embedding of embedding not flatten the structure as much as possible? Am I actually getting multiple versions, and go is not doing what I expect. I am pretty confused and would say not to merge unless we understand this correctly, as potentially other things do not work as expected in this change.

maker/widget_maker.go
```diff
@@ -168,10 +168,10 @@ func MakeWidget(
 		widget = spotifyweb.NewWidget(app, pages, settings)
 	case "status":
 		settings := status.NewSettingsFromYAML(widgetName, moduleConfig, globalConfig)
-		widget = status.NewWidget(app, settings)
+		widget = status.NewWidget(app, pages, settings)
 	// case "system":
 	// 	settings := system.NewSettingsFromYAML(widgetName, moduleConfig, globalConfig)
-	// 	widget = system.NewWidget(app, date, version, settings)
```
senorprogrammer: I've taken the settings widget out because I really didn't like how different the signature was, and the info is not very useful to end-users.

Seanstoppable:
I'm wondering if I am going about this backwards. Rather than trying to compose at this level (ala Java/Scala composable), I could pass some of these as parameters. This would, for example, let us have a MultiSource/Scrollable widget by:

Creating a TextWidget (might leave keyboard here)
Pass this TextWidget pointer to Scrollable (sets the appropriate stuff for keyboard, etc)
Pass this TextWidget pointer to MultiSourceWidget (sets the appropriate stuff for keyboard, etc)

And then if these are embedded in our final widget, we have the result without having to create a specific thing for multisourcescrollable.

senorprogrammer:
What you describe above sounds more flexible to me. When I saw this:
```
widget.ScrollableWidget.TextWidget.KeyboardWidget.SetView(widget.View)
```
it didn't feel right though I couldn't say specifically why other than it seems like forcing an inheritance approach. An interface-based approach would be better. For instance, `ScrollableWidget` should be able to accept a `TextWidget`, or anything that conforms to a `Scrollable` interface which defines the funcs necessary for working with a `ScrollableWidget`.

PR #450, opened 2019-05-13
Make github a multisource widget

Seanstoppable:
Note, that this is a backwards incompatible change

Previous config:

```
repositories:
  wtf: "wtfutil"
```

New Config:

```
repositories:
- "wtfutil/wtf"
```

### DIFF
--- a/modules/github/settings.go
+++ b/modules/github/settings.go
@@ -1,6 +1,7 @@
 package github
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/olebedev/config"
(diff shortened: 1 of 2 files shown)


senorprogrammer:
It doesn't have to be backwards incompatible. There's already an established precedent for changing config types. See how `opsgenie` handles `schedule`.

 In this case you'd first try to cast to `UList`. If that works then no change necessary. If that raises an error, then try to cast to `UMap`. If that doesn't raise an error, concat all the map's values and keys together (ie: "wtf": "wtfutil" becomes "wtfutil/wtf") and create the list of those.

Seanstoppable:
Thanks! Will take a look

Seanstoppable:
There is one major difference here in that MultiSourceWidget expects a specific config type to get the sources.
However, like how scrollable items were being handled, it was really only trying to use them to get the length, and I can refactor multisourcewidget to update that length vs reading config

senorprogrammer:
I'm not sure what you mean. It looks like Multisource reads the `len()` of its `Sources` attribute, which is `[]string`. That can still be built that way during the settings loading. Or am I missing something?

Seanstoppable:
You are correct, I am now handling both configs and works fine by setting Sources manually in construction.

senorprogrammer:
Thanks!

PR #453, opened 2019-05-13
Flip todoist over to multiview widget

Seanstoppable:
This can also probably be made scrollable in the future as well

### DIFF
--- a/wtf/utils.go
+++ b/wtf/utils.go
@@ -149,7 +149,7 @@ func ToStrs(slice []interface{}) []string {
 	results := []string{}
 
 	for _, val := range slice {
-		results = append(results, val.(string))
+		results = append(results, fmt.Sprintf("%v", val))
 	}
 
 	return results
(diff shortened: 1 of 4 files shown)


wtf/utils.go
```diff
@@ -149,7 +149,7 @@ func ToStrs(slice []interface{}) []string {
 	results := []string{}
 
 	for _, val := range slice {
-		results = append(results, val.(string))
+		results = append(results, fmt.Sprintf("%v", val))
```
Seanstoppable: This exists because I was getting a type assertion error for turning an int (todoist projects are ints) into a string. Since `%v` prints the object value, this seemed the easiest thing to do to handle strings and ints without error handling or reflection.

senorprogrammer: Something like this should work and is extensible for other types should they come up:
```
for _, val := range slice {
		switch val.(type) {
		case int:
			results = append(results, strconv.Itoa(val.(int)))
		case string:
			results = append(results, val.(string))
		}
	}
```

senorprogrammer:
👍

PR #458, opened 2019-05-19
remove glog mod dep

bosr:
Hi, I think there is a problem regarding that `glog` commit in `go.mod`: 
- `v0.0.0-20170312005925-543a34c32e4d`
it does not exist on github.

Removing this dep allows the modules retrieval to finish and the build to start (else the build fails on various CI). Btw, do you know why this dep was added in the latest commit?

Thanks!

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,6 @@ require (
 	github.com/go-test/deep v1.0.1 // indirect
 	github.com/godbus/dbus v5.0.1+incompatible
 	github.com/gogo/protobuf v1.2.1 // indirect
-	github.com/golang/glog v0.0.0-20170312005925-543a34c32e4d // indirect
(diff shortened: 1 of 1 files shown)


bosr:
If this is merged, could you please create a new release? We are trying to get this formula on Homebrew, but we need a release. Thanks!

senorprogrammer:
I don't know why this specifically was added. A assume a transitive dependency from an existing package that was updated, but I'm not sure which.

senorprogrammer:
WTF v0.10.3 has been released https://github.com/wtfutil/wtf/releases/tag/v0.10.3

Having this in Homebrew would be awesome, thanks for your effort!

bosr:
Thanks for the comment and for being so reactive! I'm just helping a new contributor with the formula. 
Keep up the good work, I love this tool, cheers!

PR #460, opened 2019-05-21
Some code improvements

sosiska:
* Some assignments simplified by using assignment operators
* Rewrite switch statement with only one case as if.
* Rewrite if-else chain as a switch statement. Based on Go style
guide: https://golang.org/doc/effective_go.html#switch
* go fmt `modules/todoist/project.go` file.


senorprogrammer:
Nice work, thanks!

PR #462, opened 2019-05-24
Migrate gitlab to multisource widget

Seanstoppable:


Seanstoppable:
Granted, I don't have a gitlab account to test with, but this seems straightforward given other migrations.

senorprogrammer:
👍

PR #464, opened 2019-05-24
Add more 'man page' like functionality

Seanstoppable:


Seanstoppable:
This puts us on a path to autogenerating docs for widgets.
Aimed initially at the command line level, but presumably we can eventually extend this to generate the markdown for wtfdocs

PR #472, opened 2019-06-04
feat(newrelic): Display deploy information from multiple apps

### LINKED ISSUE #471: Allow multiple applications to be monitored in NewRelic
Currently only one applications deploys can be tracked, where most users of NewRelic will have multiple.

Similar to other widgets this should allow for multiple "pages" to be tabbed through, one per application registered. Config would be updated to allow an array of application IDs rather than a single value.

EDIT: Meant to add - my knowledge of golang is very minimal, but I'm happy to try and tackle this if the feature is generally agreed upon.

ChrisDBrown:
Closes #471

There's a breaking change here in the config options - `applicationID` becomes `applicationIDs` and expects an array instead of int.

Docs update for the changes here: https://github.com/wtfutil/wtfdocs/pull/38

As mentioned on the original issue, I'm pretty green at golang so feedback is very welcome!


modules/newrelic/keyboard.go
```diff
@@ -0,0 +1,9 @@
+package newrelic
+
+import "github.com/gdamore/tcell"
+
+func (widget *Widget) initializeKeyboardControls() {
+	widget.SetKeyboardChar("/", widget.ShowHelp, "Show/hide this help window")
```
Seanstoppable: Can you take a look at the JIRA module and add appropriate keyboard commands? `r/j/k` would be appreciated for consistency.

modules/newrelic/settings.go
```diff
@@ -12,19 +12,19 @@ const defaultTitle = "NewRelic"
...
 
 	settings := Settings{
 		common: cfg.NewCommonSettingsFromModule(name, defaultTitle, ymlConfig, globalConfig),
 
-		apiKey:        ymlConfig.UString("apiKey", os.Getenv("WTF_NEW_RELIC_API_KEY")),
-		applicationID: ymlConfig.UInt("applicationID"),
-		deployCount:   ymlConfig.UInt("deployCount", 5),
+		apiKey:         ymlConfig.UString("apiKey", os.Getenv("WTF_NEW_RELIC_API_KEY")),
+		deployCount:    ymlConfig.UInt("deployCount", 5),
+		applicationIDs: ymlConfig.UList("applicationIDs"),
```
Seanstoppable: You could keep applicationID for a non-breaking change. Either by appending it to applicationIDs, or doing something like the `arrayifyProjects` in the jira settings file

modules/newrelic/widget.go
```diff
@@ -1,93 +1,69 @@
...
 
 func (widget *Widget) Refresh() {
-	app, appErr := widget.client.Application()
-	deploys, depErr := widget.client.Deployments()
-
-	appName := "error"
-	if appErr == nil {
-		appName = app.Name
+	for _, id := range wtf.ToInts(widget.settings.applicationIDs) {
+		widget.Clients = append(widget.Clients, NewClient(widget.settings.apiKey, id))
```
Seanstoppable: This will create a new client that is calling the service each time, and this is probably not quite what you want. Say you had 3 applications, on the first refresh there would be 3, second would be 6, third would be 9..... eventually this will eat up a lot of resources.
It might make more sense to have the client accept applicationId as a parameter, and use that to iterate through requests and store the results in some struct, effectively caching the results. Otherwise, every time you change a page you will ALSO be making a web request to the newrelic API to get data, resulting in slower transitions. Minimally, I would create one set of clients up front, and Refresh will just call the display method

senorprogrammer:
@ChrisDBrown  If there's no feedback from you by Aug 4 on the comments in this PR I'll assume it's unmaintained and will close it out as such.

Seanstoppable:
I'm happy to refactor some of this to address my comments, either as part of this PR, or opening a new one. This IS most of the way done it seems. Though I don't have a new relic account to test

senorprogrammer:
Added via https://github.com/wtfutil/wtf/pull/607

PR #483, opened 2019-07-01
Allow people to substitute a title for cmdrunner

Seanstoppable:
This falls back to existing logic otherwise


senorprogrammer:
Sweet, thanks!

Issue #481, opened 2019-07-01
Unable to display cryptolive


askemeline:
Hi, I'm not getting cryptolive value, but I'm getting others modules very well. Anyone got the same problem ? thanks

senorprogrammer:
Looks like this module is currently broken.

Seanstoppable:
I'm debugging and should have a PR soonish

senorprogrammer:
Fixed

PR #487, opened 2019-07-04
Fix cryptolive widget

Seanstoppable:
Addresses #481
Inverts the item list return, to make sure we always trigger wg.Done and not hang the widget
Clean up config for toplist widget, as we were assigning to `currencies` rather than `top` erroneously


senorprogrammer:
Very nice, thanks!

PR #488, opened 2019-07-04
Even more help settings

Seanstoppable:
Blitz through g
Clean up a setting that no longer exists for gcal


senorprogrammer:
🎉

senorprogrammer:
Starting to think about how to generate all the documentation programmatically, with these as the foundation.

Issue #482, opened 2019-07-01
Missing position data kills the app with os.Exit(1) and no useful information


senorprogrammer:
The following two formats for invalid position data cause the app to die with no useful debugging information:

```
position:
top: 1
left: 1
width: 1
height: 1
```
and 
```
# position:
    # top: 1
    # left: 1
    # width: 1
    # height: 1
```
In both cases useful debugging information should be provided.

senorprogrammer:
Closed by https://github.com/wtfutil/wtf/pull/489

PR #489, opened 2019-07-05
 WTF-482 Gracefully handle missing position configuration

senorprogrammer:
Fix for https://github.com/wtfutil/wtf/issues/482

### DIFF
--- a/cfg/position.go
+++ b/cfg/position.go
@@ -21,7 +21,7 @@ type Position struct {
 }
 
 // NewPositionFromYAML creates and returns a new instance of Position
-func NewPositionFromYAML(name string, moduleConfig *config.Config) Position {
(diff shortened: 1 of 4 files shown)


cfg/position.go
```diff
@@ -0,0 +1,92 @@
...
+type Position struct {
+	Height int
+	Left   int
+	Top    int
+	Width  int
+}
+
+// NewPositionFromYAML creates and returns a new instance of Position
+func NewPositionFromYAML(moduleName string, moduleConfig *config.Config) Position {
+	errs := make(map[string]error)
```
senorprogrammer: Make this a struct that holds the name, the error, and the current value so the current value can be displayed in the error message.

Issue #382, opened 2019-02-11
Need timezone in Google Calendar widget


audioboxer217:
#### What's the problem?
The "Family" calendar in Google Calendar is hard-coded to UTC timezone and cannot be changed.  It depends on the client's Timezone to convert appropriately.  Since WTF doesn't have a timezone it assumes the same UTC and thus all my times for my Family calendar appear incorrect.

#### How did you expect it to behave?
I would expect it to convert the timezone appropriately.  Ideally, it would use the system's timezone but I'm perfectly fine with this just being a setting in the config.

#### Operating system 
OS X Mojave (10.14.2)
Ubuntu 18.10

#### Terminal program & version
iTerm
Terminal (default in OS X)
Terminal (default in Ubuntu)

#### WTF version
0.4.0

audioboxer217:
Screenshot as an example.  Notice "Gymnastics Practice" is at 16:00 my local time (US Central) which shows in Google Calendar but WTF is showing 22:00 which is the UTC equivalent.
In WTF:
<img src="https://user-images.githubusercontent.com/6342499/52588534-4ee5b500-2e02-11e9-83f1-d6c991583447.png" width="400" height="600">

In Google Calendar:
![image](https://user-images.githubusercontent.com/6342499/52588650-92402380-2e02-11e9-9aba-3d604465b198.png)

senorprogrammer:
Fixed by @jeangovil in https://github.com/wtfutil/wtf/pull/490

PR #490, opened 2019-07-05
add time zone support to google calendar module

jeangovil:
Fixes issue #382.
Now we're able to change the time zone. The default value remains the same as before, the one set in your Google calendar settings.

### DIFF
--- a/modules/gcal/client.go
+++ b/modules/gcal/client.go
@@ -56,8 +56,10 @@ func (widget *Widget) Fetch() ([]*CalEvent, error) {
 	startTime := fromMidnight().Format(time.RFC3339)
 	eventLimit := int64(widget.settings.eventCount)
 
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Awesome. Is there somewhere that shows what a valid timezone string would look like that can be either linked to in the `help` text or act as an example in the `help` text & documentation? I suspect most people won't know what to use.

Would a valid value be anything from the `TZ database name` column here? https://en.wikipedia.org/wiki/List_of_tz_database_time_zones

jeangovil:
@senorprogrammer Yes. The values from `TZ database name` column are valid.

senorprogrammer:
Awesome, thanks!

senorprogrammer:
Note that I've switched the config attribute name from "timeZone" to "timezone". I think that will be the more expected spelling.

PR #492, opened 2019-07-07
WIP Reimpliments the module configuration validation

senorprogrammer:
Now supports displaying errors from multiple widgets.

The only `Common` setting that doesn't have logical defaults is `Position`, so this only improves things for those values.

However it sets up the possibility of having per-module configuration validation for module-specific settings that can't have logical defaults (like `apiKey`, or `username`, for instance).


senorprogrammer:
Screen Shot 2019-07-07 at 4 08 16 PM

PR #513, opened 2019-07-21
Fix rollbar rendering

### LINKED ISSUE #507: Rollbar module broken?
## What's broken?
Hello, been trying to configure the rollbar module. Nothing gets displayed, and after adding some logs, it looks like the refreshInterval parameter is not being configured.

Shouldn't the last call from widget.Refresh not be to Refresh, but perhaps widget.Render instead?

Thanks

Seanstoppable:
We should be calling Render after a refresh, not calling refresh for an infinite loop
Fixes #507


senorprogrammer:
Thanks!

PR #516, opened 2019-07-23
Added to opsgenie module region selection

l13t:
According to OpsGenie API documentation:

`If using the EU instance of Opsgenie, the URL needs to be https://api.eu.opsgenie.com for requests to be successful.`

P.S.
I'm not developer but I tried to do all my best to do this change :)

### DIFF
--- a/modules/opsgenie/settings.go
+++ b/modules/opsgenie/settings.go
@@ -13,7 +13,7 @@ type Settings struct {
 	common *cfg.Common
 
 	apiKey                 string   `help:"Your OpsGenie API token."`
-	isEurope               bool     `help:"Defines if European servers should be used." optional:"true"`
+	region                 string   `help:"Defines region to use. Possible options: us (by default), eu." optional:"true"`
 	displayEmpty           bool     `help:"Whether schedules with no assigned person on-call should be displayed." optional:"true"`
 	schedule               []string `help:"A list of names of the schedule(s) to retrieve."`
 	scheduleIdentifierType string   `help:"Type of the schedule identifier." values:"id or name" optional:"true"`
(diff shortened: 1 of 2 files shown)


modules/opsgenie/settings.go
```diff
@@ -13,6 +13,7 @@ type Settings struct {
 	common *cfg.Common
 
 	apiKey                 string   `help:"Your OpsGenie API token."`
+	isEurope               bool     `help:"Defines if European servers should be used." optional:"true"`
```
Seanstoppable: We could also consider an optional api url, which some other modules have. This means that if OpsGenie sets up another one, this module will just immediately support it.

senorprogrammer: Ideally people wouldn't have to know the API URN to use, that's a point of friction that we can abstract away (I believe the other cases where they do, those are for custom installations of services).

I like this approach for now, with one suggestion: changing it to `useEuropeanServiceRegion`. This fits with the naming that OpsGenie uses in their documentation.

Thoughts?

Seanstoppable: Fair point. If we want to abstract, I would consider abstracting to a region, with the region defaulting to US over a boolean. Better backwards compatibility if they expand and, if not, sufficiently understandable verbosity.

senorprogrammer: I like the sound of that. Extensible without the potential proliferation of booleans.
```
    opsgenie:
      apiKey: "abc-123"
      region: ["america" | "europe"]
      ...
```
@l13t you say you're not a developer. You've done a great job so far. Are you comfortable implementing this addition, or would you prefer one of us tackle it?

l13t: @senorprogrammer, sure. I'll rewrite the code as you suggested. I thought about such a way but was not sure about how to limit the number of choices in the variable (to have only `us` and `eu` for now.).

l13t:
@senorprogrammer new version pushed.

Seanstoppable:
This looks great!

senorprogrammer:
This will be merged in today.

PR #518, opened 2019-07-25
Explicitly set the default config file's mode to 0600

senorprogrammer:


cfg/error_messages.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+/* -------------------- Unexported Functions -------------------- */
+
+func displayError(err error) {
+	fmt.Printf("%s %s\n\n", aurora.Red("Error:"), err.Error())
+}
+
+func displayDefaultConfigWriteError(err error) {
+	fmt.Printf("\n%s Could not write the default configuration.\n", aurora.Red("ERROR:"))
```
Seanstoppable: I think this is a little obtuse, considering we throw it when permissions are incorrect. We should explicitly call out permissions here I think?

senorprogrammer: This error will happen if the default `config.yml` cannot be written for any reason into `~/.config/wtf/` on first create.

Issue #520, opened 2019-07-29
Esc key does not dismiss modal dialog in Todo


senorprogrammer:
The Esc key should dismiss the modal dialog in the Todo module. It no longer does.

senorprogrammer:
Tab key should also move between fields/buttons. No longer does.

Seanstoppable:
As far as I can tell, the top level keybindings are stopping widget level keybindings. Potentially a problem with the way I restructured some of the code. I'll try to dig in this weekend.

Seanstoppable:
It actually has to do with this code block:
https://github.com/wtfutil/wtf/blob/master/app/wtf_app.go#L83-L95

The 'return nil' means that the key doesn't flow through to the widgets themselves.

Seanstoppable:
Opened a PR where the behavior appears to be correct (at least for TODO). Tab doesn't cycle through when the window captures it. Esc seems to work appropriately

senorprogrammer:
Fixed by https://github.com/wtfutil/wtf/pull/523

PR #523, opened 2019-08-04
Fix keyboard fallthrough

Seanstoppable:
We are expecting Esc and Tab to perform other actions, so these need to not return nil
Addresses #520


senorprogrammer:
Awesome. I love a fix that is all deletion

Issue #142, opened 2018-06-03
Kubernetes Module


spaghettifunk:
### Feature Request
It would be nice to have a module to get the information about the resources in a kubernetes cluster.
Seeing the pods, deployments and such would be very nice but also, it would be rather cool to see the metrics of the resources regarding the nodes.

#### Summary
It should fairly simple to connect to the cluster by pointing to the `kubeconfig` file. Formatting when there are a lot of pods though, is quite a challenge I believe.

steve-winter:
I am looking at this, and now have a version that can present Kubernetes info.

My view is it may be more relevant to just highlight counts of Pods in each state, and possibly wider resource information. Something like;
```
Namespace: my-namespace
Healthy Pods: 23
Unhealthy Pods: 2 (Everything that is not ready)
```
Uses k8s.io APIs which has added ~8mb in vendoring to project so far.

senorprogrammer:
Added by @sudermanjr in https://github.com/wtfutil/wtf/pull/526

PR #526, opened 2019-08-06
Kubernetes Addon

sudermanjr:
I took a pass at implementing #142.  This kubernetes module will allow you to watch nodes, deployments, and pods in any number of namespaces.  Different output assumptions are made depending on the combination of options.


senorprogrammer:
This is great, nice work! Travis seems to be failing because of `testify`, which is no fault of yours.

modules/kubernetes/widget.go
```diff
@@ -0,0 +1,209 @@
...
+
+	client := widget.getInstance()
+
+	var content string
+
+	// Debug Info
+	//content += fmt.Sprintf("Namespaces: %q %d\n", widget.namespaces, len(widget.namespaces))
+	//content += fmt.Sprintf("Objects: %q %d\n\n", widget.objects, len(widget.objects))
+
+	if !utils.DoesNotInclude(widget.objects, "nodes") {
```
senorprogrammer: I really should add an `Includes()` function....

modules/kubernetes/widget_test.go
```diff
@@ -0,0 +1,58 @@
...
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			widget := &Widget{
+				title:      tt.fields.title,
+				namespaces: tt.fields.namespaces,
+			}
+			assert.Equal(t, tt.want, widget.generateTitle())
+		})
+	}
+}
```
senorprogrammer: Sweet. I think this might be the first test written inside the `modules/` dir. I like it a lot.

senorprogrammer:
I'll try to get this merged in tomorrow.

modules/kubernetes/widget.go
sudermanjr: I _almost_ wrote one instead of doing this, but I got lazy. :-D

modules/kubernetes/widget_test.go
sudermanjr: Thanks!

senorprogrammer:
@sudermanjr I'm writing the documentation for this module. Could you share an example of config for it? Redacting any sensitive info of course.

senorprogrammer:
Just noticed you added the `kubernetes_config.yml` sample config file. That's fantastic, thank you.

Issue #529, opened 2019-08-11
wtfutil panics with SIGSEGV on module info


iNode:
## What's broken?
wtfutil crashes on request to get module info as per app usage. Binary is downloaded from the release page of repository. OS version - Ubuntu 18.04.

Maybe todo module is not available anymore and/or usage might be not up to date, but likely it should not panic.

```
$ wtfutil -v
0.19.1

$ wtfutil -h
Usage:
  wtfutil [OPTIONS]

Application Options:
  -c, --config=  Path to config file
  -m, --module=  Display info about a specific module, i.e.: 'wtf -m=todo'
  -p, --profile  Profile application memory usage
  -v, --version  Show version info

Help Options:
  -h, --help     Show this help message

$ wtfutil -m=todo
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x57c537]

goroutine 1 [running]:
github.com/olebedev/config.(*Config).Bool(0x0, 0x120d54f, 0x7, 0xc0001e1600, 0xc000039d00, 0xd)
        /Users/ccummer/go/pkg/mod/github.com/olebedev/config@v0.0.0-20190528211619-364964f3a8e4/config.go:147 +0x37
github.com/olebedev/config.(*Config).UBool(0x0, 0x120d54f, 0x7, 0xc00054969f, 0x1, 0x1, 0xd)
        /Users/ccummer/go/pkg/mod/github.com/olebedev/config@v0.0.0-20190528211619-364964f3a8e4/config.go:162 +0x3f
github.com/wtfutil/wtf/maker.MakeWidget(0x0, 0x0, 0x7fffe6a0d34f, 0x4, 0xc0001e4080, 0x40c478, 0x20)
        /Users/ccummer/go/src/github.com/wtfutil/wtf/maker/widget_maker.go:68 +0xe5
github.com/wtfutil/wtf/help.helpFor(0x7fffe6a0d34f, 0x4, 0xc0001e4080, 0x0, 0x0)
        /Users/ccummer/go/src/github.com/wtfutil/wtf/help/help.go:21 +0x4b
github.com/wtfutil/wtf/help.Display(0x7fffe6a0d34f, 0x4, 0xc0001e4080)
        /Users/ccummer/go/src/github.com/wtfutil/wtf/help/help.go:16 +0xb4
[...]

PR #530, opened 2019-08-12
Fix help for modules

Seanstoppable:
This does 2 things:
1. Forces a module to be 'enabled' on the help route, so that it works
2. Better handling when no module is actually found

Addresses #529

### DIFF
--- a/help/help.go
+++ b/help/help.go
@@ -9,16 +9,23 @@ import (
 )
 
 // Display displays the output of the --help argument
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Awesome. @seanstoppable I won’t be in front of a computer for another week. I think you have merge capability. If so, please feel free

Seanstoppable:
Looks like I don't, but no worries.

senorprogrammer:
You should now.

Issue #534, opened 2019-08-19
Power module's "Remaining" makes no sense when on A/C


senorprogrammer:
## What's broken?

When on AC power, the "Remaining" field says "0:00". Should say something more sensible (like ∞ maybe).

Seanstoppable:
Is this when fully charged? When charging, I see `∞` on both OSX and Ubuntu

PR #535, opened 2019-08-21
Fix fully charged indicator

Seanstoppable:
pmset shows `0:00` when charged. Nothing when charging/discharging
Handle this edge case


senorprogrammer:
Thanks! Closes #534

PR #536, opened 2019-08-21
Attempt to offer linux transparency

Seanstoppable:
Add 'transparent' as a color option
Default to 'transparent'
Make sure to set color to appropriate things, so transparency is achieved


senorprogrammer:
Having `tcell.ColorDefault` be the defaults is a great improvement.

senorprogrammer:
Thanks!

PR #538, opened 2019-08-25
Fix a bunch of race conditions

Seanstoppable:
GetWidth causes a race condition

Migrate to a function called within `QueueUpdateDraw`

After these changes, my setup didn't trigger a race condition despite running for a few hours

### DIFF
--- a/modules/datadog/widget.go
+++ b/modules/datadog/widget.go
@@ -60,8 +60,7 @@ func (widget *Widget) Refresh() {
 }
 
 func (widget *Widget) Render() {
-	content := widget.contentFrom(widget.monitors)
-	widget.Redraw(widget.CommonSettings().Title, content, false)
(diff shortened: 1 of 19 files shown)


senorprogrammer:
Looks solid. 

A couple questions about this one. This only changes redraw in some of the modules, not all. Is there a reason the other modules don't need this?

In `view/text_widget.go` we now have both `func Redraw()` and `func RedrawFunc()` as exported methods that do effectively the same thing. Can `Redraw()` be removed?

Seanstoppable:
The other modules don't "need" this since they don't call unsafe methods. I'll update everything and remove Redraw for consistency though.

Seanstoppable:
I'd actually like to refactor the other ones in smaller chunks. In particular ScrollableWidgets look like they may be a bit of work

PR #555, opened 2019-08-29
Add README instructions on installing via MacPorts

herbygillot:
You can now install wtfutil via MacPorts.


senorprogrammer:
This is fantastic, thanks!

PR #559, opened 2019-08-29
add FreeBSD support to power module

hxw:
uses the FreeBSD apm command to get charge status
and battery percentage

Signed-off-by: Christopher Hall <hsw@ms2.hinet.net>


senorprogrammer:
Thanks for this, this is great.

PR #570, opened 2019-08-30
add arpansagovau module for displaying UV data from Australian cities

jeffz:


modules/weatherservices/arpansagovau/settings.go
```diff
@@ -0,0 +1,22 @@
...
+
+type Settings struct {
+	common *cfg.Common
+	city string;
+}
+
+func NewSettingsFromYAML(name string, ymlConfig *config.Config, globalConfig *config.Config) *Settings {
+	settings := Settings{
+		common: cfg.NewCommonSettingsFromModule(name, defaultTitle, ymlConfig, globalConfig),
+		city: ymlConfig.UString("locationid"),
```
senorprogrammer: What happens if there is no city defined? Does it make sense to default it to something?

senorprogrammer:
This looks great, it'll be included in the next release.

modules/weatherservices/arpansagovau/settings.go
jeffz: Good question.  If no city is defined, the widget displays "no data?" in red.   

As for a default, I don't feel too strongly about setting one.   

The accompanying PR for wtfdocs details a suggested default "Sydney" fwiw for the prospective user: https://github.com/wtfutil/wtfdocs/pull/53/commits/a79b9d87525abe30a48e3ba9fdeee9a03de8c9be - which I think is more useful to know than to have a default.

senorprogrammer:
Thanks!

PR #576, opened 2019-08-30
Google Calendar Read Level

mikkeljuhl:
This pull request creates a new environment varialble `calendarReadLevel`.

**The reason for this change:** If environment variable is set to `writer` calendars you subscribe to will not show up. However, if the variable is set to `reader` then both your calendars and calendars you subscribe to will show up.

This change keeps the `writer` option as default. 

Further justification: If you subscribe to a calendar, for instance a work calendar, school calendar, then it will not show up with the `writer` role. It will however show up using the `reader` role.


senorprogrammer:
Thanks!

PR #580, opened 2019-08-30
Add more configuration options to travisci module

nyourchuck:
I ran into an issue with the list of builds pulled from https://developer.travis-ci.com/resource/builds#Builds

By default travis sorts the builds by "number:desc,id:desc", which meant that the widget only displayed builds from a 2016 project which had a build number over 1800. Changing the sort to "id:desc" shows the most recently created builds. It is also possible for users to configure their preferred sort fields and the number of builds to display in the widget.

I'll also update the docs repository if these changes are approved.

### DIFF
--- a/modules/travisci/client.go
+++ b/modules/travisci/client.go
@@ -15,12 +15,12 @@ var TRAVIS_HOSTS = map[bool]string{
 	true:  "travis-ci.com",
 }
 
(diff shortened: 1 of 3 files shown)


senorprogrammer:
This looks good. An update to the docs repo would be appreciated so they can be merged in unison.

senorprogrammer:
Thanks!

PR #584, opened 2019-08-31
Find SSID regardless of network manager

alexfornuto:
I made this change on my own system, where I used wicd instead of NetworkManager. I _think_ that the wireless-tools package containing `iwgetid`is a standard package on most systems, so this change would potential make the security module work across more Linux systems.

#hacktoberfest

### DIFF
--- a/modules/security/wifi.go
+++ b/modules/security/wifi.go
@@ -66,11 +66,10 @@ func wifiInfo() string {
 }
 
 func wifiNameLinux() string {
-	cmd := exec.Command("nmcli", "-t", "-f", "in-use,ssid", "dev", "wifi")
-	out := utils.ExecuteCommand(cmd)
-	name := utils.FindMatch(`\*:(.+)`, out)
+  cmd := exec.Command("/usr/sbin/iwgetid", "-r")
+	name := utils.ExecuteCommand(cmd)
 	if len(name) > 0 {
-		return name[0][1]
+		return name
 	}
 	return ""
 }
--- a/modules/security/wifi.go
+++ b/modules/security/wifi.go
@@ -66,13 +66,11 @@ func wifiInfo() string {
 }
 
 func wifiNameLinux() string {
-  cmd := exec.Command("/usr/sbin/iwgetid", "-r")
-	name := utils.ExecuteCommand(cmd)
-	if len(name) > 0 {
-		return name
-	}
-	return ""
+  cmd, _ := exec.Command("iwgetid", "-r").Output()
+  return string(cmd)
 }
+
+
 
 func wifiNameMacOS() string {
 	name := utils.FindMatch(`s*SSID: (.+)s*`, wifiInfo())


modules/security/wifi.go
```diff
@@ -66,11 +66,10 @@ func wifiInfo() string {
 }
 
 func wifiNameLinux() string {
-	cmd := exec.Command("nmcli", "-t", "-f", "in-use,ssid", "dev", "wifi")
-	out := utils.ExecuteCommand(cmd)
-	name := utils.FindMatch(`\*:(.+)`, out)
+  cmd := exec.Command("/usr/sbin/iwgetid", "-r")
```
alexfornuto: I'm not sure why this is showing as indented differently, it looks the same in vim **and** GitHub's editor.

senorprogrammer: Is this utility guaranteed to always be at that absolute path on all systems?

senorprogrammer:
This is a potentially breaking change and so far there have been no complaints about the use of `nmcli`. This also seems to conflict with #585, which has similar changes. What's the expectation here?

alexfornuto:
> This is a potentially breaking change 
True. So far all the systems I've tested, (Ubuntu, Debian, Arch Linux) have had `iwgetid`.

>so far there have been no complaints about the use of nmcli.
Consider this the first complaint. nmcli is a NetworkManager tool, and those using other networking services won't be able to use it.

> This also seems to conflict with #585, which has similar changes.
#585 is also my PR, and says that it's dependent on this PR - it doesn't conflict, it continues based on this patch.


> What's the expectation here?

>>[so this change would potential make the security module work across more Linux systems.](#issue-312941392)

senorprogrammer:
The issue I have is that this is a breaking change and the next PR says not to merge it in. So as it stands right now, I don't see either of these going into `master` as-is.

I would support being able to switch between `nmcli` and `iwgetid` using a config flag for Linux on the `security` module.

alexfornuto:
Sorry if the language in #585 isn't clear, but it says that _it_ should not be merged as is. But you're right about everything else.

When I have the time I'll try to learn how to make the changes you mentioned to to the security modules. In the meantime, if you can link me to any recommended documentation on how to add the flags in Golang, I'd appreciate it.

alexfornuto:
In fact, I think I might try an if statement to see if `nmcli` works, and fallback to my solution if not.

senorprogrammer:
@alexfornuto That sounds good to me

modules/security/wifi.go
alexfornuto: I'm not sure yet, but calling it without absolute path failed for me.

senorprogrammer:
@alexfornuto Any luck with the `if` statement approach? I don't want to merge this in without understanding the impact it might have on other Linux users (unfortunately I don't have a Linux box to test this on)

alexfornuto:
@senorprogrammer I haven't had a chance to try yet. If you want to close this out until I get some Go experience, I'll understand and re-open/re-submit at a later date.

alexfornuto:
@senorprogrammer et al: I've updated this PR to be more awesome:

 - It no longer calls `iwgetid` from the full path. I thought it did because of weirdness where packages in `/usr/sbin` don't always tab complete in shells.
 - It now properly throws away error statuses when there is no network connection. This also reduces complexity in the code because we don't need to check the length of the output. It's either empty, or the network name.

Regarding concerns that this would be a breaking change, [Wikipedia](https://en.wikipedia.org/wiki/Wireless_tools_for_Linux#Adoption) confirms my original thinking:

>It is included with most operating system distributions built on the Linux kernel. In many Linux distributions, this package is included by default, or based on whether a wireless card is present. If it is not automatically installed by the distribution, it is usually easy to find in binary form.

This refers to the `wireless-tools` meta package, which includes `iwgetid`.

At the moment this module is still dependent on NetworkManager to find the crypto status, but I cna work on that in a separate issue. I consider this one ready for review.

P.S. As an eternal contributor, I am unable to tag my own PR with the "Hacktoberfest" label.

P.P.S. If you're still worried about breaking, I can (try to) write you a CircleCi test to check this against different distros.

senorprogrammer:
Thanks for all this, I will look into it this weekend and try to get it into master.

senorprogrammer:
Works on my RaspPis. 👍

nullvariable:
Can confirm this is working for me in Ubuntu 18.04

PR #585, opened 2019-08-31
Wicd

alexfornuto:
This depends on #584 

This change pulls the Wireless encryption method from wicd instead of nmcli. It **Should Not** be merged as is, since many more systems are using NetworkManager than wicd.

I am not a developer, but I figured out this much. I'm hoping that some generous soul will ammend this PR with the Go code making this an elif statement such that if the nmcli output is a failure, it will try this code instead.

With such an elif in place, others could further expand this section to use their networking client of choice.

### DIFF
--- a/modules/security/wifi.go
+++ b/modules/security/wifi.go
@@ -43,10 +43,10 @@ func WifiName() string {
 /* -------------------- Unexported Functions -------------------- */
 
 func wifiEncryptionLinux() string {
-	cmd := exec.Command("nmcli", "-t", "-f", "in-use,security", "dev", "wifi")
-	out := utils.ExecuteCommand(cmd)
+  cmd := exec.Command("/usr/bin/wicd-cli", "--wireless", "-d")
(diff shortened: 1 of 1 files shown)


senorprogrammer:
This is an interesting idea. I'd like other knowledgable Linux users to weigh in on the ideas in here.

dasJ:
Any reason /usr/bin is hardcoded? The path is usually in $PATH and I don't see any reason to limit us here.

alexfornuto:
IIRC that had to do with me not finding the command in either my PATH or in bash completion. Now that hacktoberfest is upon us, I hope to be looking at this again soon.

senorprogrammer:
Any thoughts on continuing this PR?

alexfornuto:
IIRC correctly what it needs is a check to see if nmcli is present. Not sure how to GO that, but I'll give it a look this weekend.

senorprogrammer:
Closing this out due to inactivity. Feel free to reopen if anything changes.

PR #586, opened 2019-08-31
WTF-547 Open selected GitHub PR

### LINKED ISSUE #547: Select a GitHub PR to be opened in a browser
## What should it do?
When using the GitHub module I would like to be able to select my PR's with the arrow keys/j or k to then open that PR in a browser when hitting the return key.

Midnight-Conqueror:
This closes #547

using the up or down arrow / j or k keys it will select the PR/issue. Once selected you can hit return to open that specific PR in a browser. It also retains opening of the repo by pressing insert. Insert probably isn't the best key for this and am open to change that.

Currently it isn't using a ScrollableWidget but rather uses highlighting of text through the TextWidget. when hitting enter the PR number will be sent used in conjunction with the current repo to determine what url to open.

This could be done through a ScrollableWidget as I'm aware I've used most of the functionality from there. DRY isn't exactly followed here and would be open to turning it into a ScrollableWidget but would need some guidance.

### DIFF
--- a/modules/github/widget.go
+++ b/modules/github/widget.go
@@ -80,7 +80,7 @@ func (widget *Widget) Prev() {
 
 func (widget *Widget) Unselect() {
 	widget.Selected = -1
-	widget.View.Highlight(strconv.Itoa(widget.Selected)).ScrollToHighlight()
+	widget.View.ScrollToBeginning()
 }
 
 func (widget *Widget) Refresh() {
--- a/modules/github/widget.go
+++ b/modules/github/widget.go
@@ -81,6 +81,7 @@ func (widget *Widget) Prev() {
 
 func (widget *Widget) Unselect() {
 	widget.Selected = -1
+	widget.View.Highlight()
 	widget.View.ScrollToBeginning()
 }
 
--- a/modules/github/widget.go
+++ b/modules/github/widget.go
@@ -76,9 +76,6 @@ func (widget *Widget) Prev() {
 
 func (widget *Widget) Unselect() {
 	widget.Selected = -1
-	if widget.DisplayFunction != nil {
-		widget.SetDisplayFunction(widget.display)
-	}
 }
 
 func (widget *Widget) Refresh() {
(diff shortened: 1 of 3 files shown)


modules/github/widget.go
```diff
@@ -87,6 +129,14 @@ func (widget *Widget) currentGithubRepo() *GithubRepo {
 	return widget.GithubRepos[widget.Idx]
 }
 
+func (widget *Widget) openPr() {
+	currentSelection := widget.View.GetHighlights()
+	if widget.Selected >= 0 && currentSelection[0] != "" {
+		url := (*widget.currentGithubRepo().RemoteRepo.HTMLURL + "/pull/" +  widget.View.GetRegionText(currentSelection[0]))
+		utils.OpenFile(url)
```
senorprogrammer: This is currently breaking on my install because the selected text is padded with empty spaces, so the URL ends up being `https://github.com/wtfutil/wtf/pull/%20472`. It would be better to use the selected index to get the repo from the widget's `GithubRepos` and then use the repo's `Number` attribute. That'll make this resistant to future styling changes.

senorprogrammer: These changes look great, however this part's going to cause issues. The selected text can be padded with spaces; on my machine the URL ends up being `https://github.com/wtfutil/wtf/pull/%20584`, which is invalid.

If there's a way to do this using `currentGithubRepo()`, then using the selected line index and calling against the repo's `PullRequests` property, that would be preferable.

Midnight-Conqueror: Sounds good to me, I'm thinking of a way I can store all the issue numbers into a slice and reference them using the Selected integer

Midnight-Conqueror:
An issue I have found is highlighting of text happens even when the widget is not focused. To reproduce. Focus on widget, press down or up arrow keys, hit escape, pressing up or down arrow keys still selects new issues. However this stops once selecting another widget. Any thoughts on ways to prevents this?

modules/github/widget.go
Midnight-Conqueror: I've fixed this in commit c21b7c32a838f339b806ff886a20a35af8c07230

senorprogrammer:
Thanks!

senorprogrammer:
Regarding the highlighting issue, that's a bug in the underlying system. A module that responds to key events will still respond to those key events if no other module has focus. Something to be fixed at a lower level than this.

PR #587, opened 2019-08-31
Capture API exceptions from todoist API

Seanstoppable:
Rather than silently swallow, capture them and use them in rendering


senorprogrammer:
Thanks!

PR #588, opened 2019-08-31
Fix delete/close of todoist

Seanstoppable:
Currently, when deleting, we jump 2 positions, since we are calling next.
However, the next item becomes the current selected position, so handle better
Properly set selected in some edge cases where we may go from a list to 0


senorprogrammer:
Looks good, nice catch.

PR #590, opened 2019-09-01
Refactor a number of widgets to display client errors

Seanstoppable:
Rather than swallowing or crashing, display appropriate errors


senorprogrammer:
Looks good

PR #592, opened 2019-09-01
Generalized TODO module

Seanstoppable:
Todo module with configurable backend.
This seperates rendering from getting data
Comes with 2 implementations, Todoist (which initial implementation is based off of) and Trello
Current Todoist widget goes away in favor of this general widget (behavior is 1:1)
Trello module still exists. Functionality changed significantly, but I can remove current implementation, which is just listing, for the new interactable implementation.


Seanstoppable:
Trello api suggests that delete should be used sparingly. The functionality in backend is to do nothing, and am wondering if we should just pull it out. Close archives tasks and removes them from lists already, causing them to disappear from the list here.

senorprogrammer:
I really like this idea a lot. I can see this folding in `Todo` as well (which I will do later).

A breaking change like this effectively needs to go out on it's own, with deprecation warnings for the existing `Todoist` module.  Ideally we'd deprecate `Todoist`, `Trello`,  and `Todo` all at the same time.

It will also need extensive documentation before it goes into master, describing the migration path from existing modules.

senorprogrammer:
Mapping Trello's `close` to `delete` seems reasonable to me. Anyone who wants to actually destroy a card can do it via Trello.

Seanstoppable:
Todoist works with current config, and I can migrate trello similarly. i.e. config will not be broken, they just get a "improved" version of the widget that works like todoist.
Planning to implement a third online service to make sure it is generalized enough, if there are any recommendations.

senorprogrammer:
That sounds excellent.

Seanstoppable:
Huh, just noticed I somehow closed this. Will re-open when I get the third implementation done.

Issue #594, opened 2019-09-03
Docker Module


v-braun:
Would be great to have a (or two separate) docker modules. 

The one I'm thinking if would show container states (running / stopped / restarting).
The other could display general docker information like space usage, version, container and image count (see [docker system command](https://docs.docker.com/engine/reference/commandline/system/))

v-braun:
I opened this issue for documentation reasons. I'm currently working on this module.

senorprogrammer:
https://github.com/wtfutil/wtf/pull/596 has been merged into `master`. Nice work @v-braun.

PR #596, opened 2019-09-03
implemented docker module

v-braun:
see issue #594 

current ui:
image


senorprogrammer:
This looks good and merged well. Could you paste an example of the config in here so I can update the docs (redact any private information first).

v-braun:
There is an example config checked in beside the module code.
see here:
``` yml

    docker:
      type: docker
      title: "docker"
      enabled: true
      position:
        top: 0
        left: 0
        height: 3
        width: 3
      refreshInterval: 1
      labelColor: lightblue

```

The only setting that is available there is an additional labelColor.
Two questions:
1.: I never used the go modules till yet. So also not sure how to merge the _go.mod_ and _go.sum_ files correctly. Will you do that? And if yes: how you do it, because it looks quite hard to merge it with all the version conflicts.

2.: Should I update the docs or will you update them also?

senorprogrammer:
I'm not sure if there's an automated way to do it, I just merge `go.mod` manually and re-generate `go.sum`.

I can do the docs for this. Thanks for the PR!

PR #607, opened 2019-09-05
feat(newrelic): Display deploy information from multiple apps

Seanstoppable:
This is a rebased/minimally fixed version of #472

Changes are:
Move client config to widget setup. This means clients are initialized once, and if config is updated, we are recreating widgets, so this is should be 'fine'.
Standard keyboard config
Should still support `applicationID` vs `applicationIDs`


Seanstoppable:
@ChrisDBrown do you mind giving this a shot with NewRelic? I made the small amount of changes from my comments that should make this reasonable. Caching responses for faster paging can always be done on top of this.

PR #608, opened 2019-09-06
Fix git ordering

Seanstoppable:
This is due to us sorting
It doesn't look like there is a need to sort here, so suggest we remove it


senorprogrammer:
Makes sense to me 👍

PR #609, opened 2019-09-06
fix: lint GitHub module, fix visual icon bug 

Midnight-Conqueror:
This PR has linted the GitHub module and fixed an issue where the icons representing the mergeable status of a PR would sometimes show random characters. This was mainly down to the way the icon was references, now referencing with unicode. I don't think terminals/go/tview likes the unicode icon \u2314 so am using \u2313 to represent a tick.

Also let me know if this is the correct kind of comments to put above exported functions

### DIFF
--- a/modules/github/display.go
+++ b/modules/github/display.go
@@ -131,7 +131,7 @@ func (widget *Widget) title(repo *GithubRepo) string {
 }
 
 var mergeIcons = map[string]string{
-	"dirty":    "[red]![white] ",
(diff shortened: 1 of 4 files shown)


Midnight-Conqueror:
Here is a picture of before and after the fix, its slightly smaller but doesn't weird out and put random characters next to the tick
wtfutil_fix_comparison

senorprogrammer:
Thanks!

Issue #549, opened 2019-08-28
Segmentation fault on kubernetes module


Draiken:
## What's broken?

wtfutil crashes when enabling the kubernetes module

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x45d55a]

goroutine 49 [running]:
k8s.io/client-go/kubernetes.NewForConfig(0x0, 0x0, 0xc000501e30, 0xe)
	/Users/ccummer/go/pkg/mod/k8s.io/client-go@v12.0.0+incompatible/kubernetes/clientset.go:335 +0x4e
github.com/wtfutil/wtf/modules/kubernetes.(*Widget).getKubeClient(0xc0004e95c0, 0xc00059c030, 0xc000507608)
	/Users/ccummer/go/src/github.com/wtfutil/wtf/modules/kubernetes/client.go:32 +0x58
github.com/wtfutil/wtf/modules/kubernetes.(*Widget).getInstance.func1()
	/Users/ccummer/go/src/github.com/wtfutil/wtf/modules/kubernetes/client.go:22 +0x42
sync.(*Once).Do(0x2fe6c18, 0xc000507638)
	/usr/local/Cellar/go/1.12.7/libexec/src/sync/once.go:44 +0xb3
github.com/wtfutil/wtf/modules/kubernetes.(*Widget).getInstance(0xc0004e95c0, 0xc0003aa240)
	/Users/ccummer/go/src/github.com/wtfutil/wtf/modules/kubernetes/client.go:19 +0x55
github.com/wtfutil/wtf/modules/kubernetes.(*Widget).Refresh(0xc0004e95c0)
	/Users/ccummer/go/src/github.com/wtfutil/wtf/modules/kubernetes/widget.go:46 +0x65
github.com/wtfutil/wtf/app.Schedule(0x1f580a0, 0xc0004e95c0)
	/Users/ccummer/go/src/github.com/wtfutil/wtf/app/scheduler.go:12 +0x49
created by github.com/wtfutil/wtf/app.(*WtfApp).scheduleWidgets
	/Users/ccummer/go/src/github.com/wtfutil/wtf/app/wtf_app.go:119 +0x81
```
kubectl version (not sure it's needed) 
```
[...]

PR #615, opened 2019-09-06
Update kubernetes client to connect to various kubernetes providers

ibaum:
* Import k8s.io/client-go/plugin/pkg/client/auth to include
authentication libraries
* Update calls to panic to include the error message

This fixes https://github.com/wtfutil/wtf/issues/549 for me using GCP. I opted to include auth, and not just gcp to support other providers without enumerating.


senorprogrammer:
Thanks!

Issue #600, opened 2019-09-04
Jenkins server restarting/crashing causes WTF to crash


Midnight-Conqueror:
## What's broken?
Sometimes we restart our Jenkins server which causes the wtf app to crash due to no response. The Jenkins module should handle the lack of response from the server gracefully, preventing the app from crashing.

Seanstoppable:
Thanks! Will take a look at some point this week.

Seanstoppable:
@Midnight-Conqueror actually, which version are you running? I recently make some changes that are in master that should surface errors better for a lot of widgets and JIRA happens to be one of them.

Seanstoppable:
Sorry, misread jenkins as jira. Jenkins still has this problem. PR upcoming

Seanstoppable:
Sorry, misread JIRA for Jenkins, but I did also make improvements to the error handling for Jenkins.

Midnight-Conqueror:
I'll update to the latest and keep an eye on how it goes today

Midnight-Conqueror:
- Running the latest WTF

Jenkins server crashed once today causing the app to crash, Jenkins server also had a slow response time once causing wtf to crash. Seems this is still an issue.

Am happy to try look into this, any directions on where to start would be amazing

Seanstoppable:
Presumably there is a stack trace. That can likely be tracked back to some problem. I was initially auditing the client code, to see if we missed/failed to handle an error. Doing another pass, I noticed that if we can't parse JSON from the response, we'll panic and crash, which is probably not expected. And it could also be that in these circumstances we get a response and not an error as `httpClient.Do` treats them. 

Looks like we have 20 panics in modules, I'll create a ticket to do an audit for a better UX.

senorprogrammer:
Given the lack of interest in this issue and the supposition of a fix in place, closing this out.

PR #619, opened 2019-09-07
Fix JSON parsing

Seanstoppable:
Issue #600 points out Jenkins module panics occasionally
I noticed that a number of places panic because of JSON parsing
Centralize JSON parsing, have it return an error, and have widgets
report the error


senorprogrammer:
🎉

PR #622, opened 2019-09-09
Disable the GOSUMDB in the compilation instructions

noxer:
With Go 1.13 the current instructions result in an error message

```
verifying github.com/golang/glog@v0.0.0-20170312005925-543a34c32e4d/go.mod: github.com/golang/glog@v0.0.0-20170312005925-543a34c32e4d/go.mod: reading https://sum.golang.org/lookup/github.com/golang/glog@v0.0.0-20170312005925-543a34c32e4d: 410 Gone
```

because Go tries to verify the module hashes against the public hash database.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,9 @@ If you want to run the build command from within your `$GOPATH`:
 # Set the Go proxy variable to GoCenter
 export GOPROXY="https://gocenter.io"
 
+# Disable the Go checksum database
+export GOSUMDB=off
+
 # Enable Go modules
 export GO111MODULE=on
 


senorprogrammer:
Sounds good to me. Any idea what would be required to make the project Go 1.13-compatible? Is it just pointing the dependencies to new hash db?

noxer:
It seems like the hashes of the modules need to be in the global hash database but this will only work for public modules, not those fetched from gocenter.io.

senorprogrammer:
Thanks!

PR #626, opened 2019-09-11
Fix config reload race

### LINKED ISSUE #422: Config auto-reload has potential race condition
#### What's the problem?

When you run `go run -race main.go` and live-reload the config, it raises the potential of a race condition. This needs to be handled.

No race.

Seanstoppable:
Race condition is around the usage of `enabled`
Wrap access in a mutex to eliminate race
Fixes #422

### DIFF
--- a/view/base.go
+++ b/view/base.go
@@ -2,6 +2,7 @@ package view
 
 import (
 	"fmt"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Beautiful

PR #629, opened 2019-09-12
Added info for using Jira with Basic-Auth

noxer:
Jira offers API calls with either an email address and API key or username and password. Added doc info for the second case. This PR does not change any functionality.


senorprogrammer:
This is great, thanks!

Issue #624, opened 2019-09-10
BarGraph colours should be customizable 


senorprogrammer:
## What should it do?

Currently BarGraph's indicator colour is hard-coded as `red`. This value should be customizable per-module.

senorprogrammer:
Fixed by https://github.com/wtfutil/wtf/pull/631

PR #631, opened 2019-09-13
Improve bargraph view so bar colors are configurable

Seanstoppable:
Inspired by #624
Update the view object so that widget developers can configure bar color

### DIFF
--- a/modules/bargraph/widget.go
+++ b/modules/bargraph/widget.go
@@ -50,8 +50,9 @@ func MakeGraph(widget *Widget) {
 		barTime = barTime.Add(time.Duration(time.Minute))
 
 		bar := view.Bar{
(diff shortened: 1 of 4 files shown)


senorprogrammer:
Specs are failing. Looks like it's missing a default value.

Seanstoppable:
Expanded/fixed the test. In the case where a value isn't set, it will use whatever other coloring is set.
Example:

Screen Shot 2019-09-13 at 7 24 44 PM

senorprogrammer:
I think it should not be rendering `[[]|||` though, the `[[]` set of braces is left over because the colour is not defined inside them, so they're rendering. `white` should probably be written in there, or the braces removed.

senorprogrammer:
👍

PR #632, opened 2019-09-13
Clean up cfg error handling a bit

Seanstoppable:
Pass in the actual file being used, rather than hardcoded `config.yaml`
Differences between two error messages are not that distinct
Centralize on one and clean up all the `isCustomConfig` tracking

### DIFF
--- a/cfg/error_messages.go
+++ b/cfg/error_messages.go
@@ -42,19 +42,8 @@ func displayWtfConfigDirCreateError(err error) {
 	displayError(err)
 }
 
-func displayWtfConfigFileLoadError(err error) {
-	fmt.Printf("\n%s Could not load '%s'.\n", aurora.Red("ERROR"), aurora.Yellow(WtfConfigFile))
-	fmt.Println()
-	fmt.Println("This could mean one of two things:")
-	fmt.Println()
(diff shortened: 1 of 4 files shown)


cfg/error_messages.go
```diff
@@ -42,19 +42,8 @@ func displayWtfConfigDirCreateError(err error) {
 	displayError(err)
 }
 
-func displayWtfConfigFileLoadError(err error) {
-	fmt.Printf("\n%s Could not load '%s'.\n", aurora.Red("ERROR"), aurora.Yellow(WtfConfigFile))
-	fmt.Println()
-	fmt.Println("This could mean one of two things:")
-	fmt.Println()
-	fmt.Printf("    1. Your %s file is missing. Check in %s to see if %s is there.\n", aurora.Yellow(WtfConfigFile), aurora.Yellow("~/.config/wtf/"), aurora.Yellow(WtfConfigFile))
```
senorprogrammer: This was differentiated because in the case of an explicitly-passed in file, the user knows which file the message is talking about.

In the case above, using the default config, the user may not know where the config file lives. This message tells them that.

senorprogrammer: If that can be maintained while getting rid of the `bool` that's passed around, that would be great.

Seanstoppable: Yeah, it works.

With default (triggered by bad yaml):

Screen Shot 2019-09-13 at 7 14 14 PM

File that doesn't exist:

Screen Shot 2019-09-13 at 7 14 31 PM

senorprogrammer: Sweet.

senorprogrammer:
Very nice.

PR #639, opened 2019-09-17
Add a focusable over-ride setting for modules

senorprogrammer:
Widgets can now override a module's `defaultFocusable` value in the module config using a `focusable` key:

```
    focusable: true || false
```

### DIFF
--- a/view/text_widget.go
+++ b/view/text_widget.go
@@ -13,9 +13,9 @@ type TextWidget struct {
 }
 
 // NewTextWidget creates and returns an instance of TextWidget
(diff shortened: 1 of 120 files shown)


view/text_widget.go
```diff
@@ -13,9 +13,9 @@ type TextWidget struct {
 }
 
 // NewTextWidget creates and returns an instance of TextWidget
-func NewTextWidget(app *tview.Application, commonSettings *cfg.Common, focusable bool) TextWidget {
```
senorprogrammer: TODO: Get rid of passing `defaultFocusable` into the constructor. Have that be a constant in the module. Use the config value first and default to the module constant.

PR #653, opened 2019-09-24
support enterprise Travis endpoints

### LINKED ISSUE #652: TravisCI module should support enterprise Travis endpoints.
## What should it do?
Currently you can only target the public Travis APIs. Like the Github module, I want to use wtf with my enterprise TravisCI environment.

See https://docs.travis-ci.com/api/#making-requests

```
Travis CI Enterprise: For projects running on a custom setup, use https://travis.example.com/api (where you replace travis.example.com with the domain Travis CI is running on).
```

I've already got it working locally, and I'm about to open up a PR.

scw007:
Closes #652

### DIFF
--- a/modules/travisci/client.go
+++ b/modules/travisci/client.go
@@ -17,6 +17,9 @@ func BuildsFor(settings *Settings) (*Builds, error) {
 	builds := &Builds{}
 
 	travisAPIURL.Host = "api." + TRAVIS_HOSTS[settings.pro]
+	if settings.baseURL != "" {
+		travisAPIURL.Host = settings.baseURL
+	}
 
 	resp, err := travisBuildRequest(settings)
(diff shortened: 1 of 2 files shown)


modules/travisci/client.go
```diff
@@ -39,22 +42,25 @@ var (
 
 func travisBuildRequest(settings *Settings) (*http.Response, error) {
 	var path string = "builds"
+	if settings.baseURL != "" {
+		travisAPIURL.Path = "/api/"
+	}
```
Seanstoppable: I would consider moving these to the settings file and simplifying this here to just be whatever settings exist.

Not a requirement and thanks for the contribution!

senorprogrammer:
Thanks!

PR #660, opened 2019-09-26
Add module to display links from a specific subreddit

lawrencecraft:
This is a new module that displays links from a specific subreddit. Much like the hackernews module (on which it's heavily based), it lets you navigate to the reddit link or the link destination.

Note that this just uses the public Reddit API. It's more targeted toward someone who wishes to easily keep an eye on something like /r/buildapcsales or the like rather than someone looking to add their front page--I haven't done any authentication or anything like that.

There will shortly be a pull request incoming for documentation. Please let me know if there's anything else I need to do.


senorprogrammer:
Thanks for this. I'm currently travelling. I'll try to get this in this weekend.

PR #675, opened 2019-10-03
Add Buildkite widget

jmks:
Hi,

I've managed to cobble together a widget that checks [Buildkite](https://buildkite.com), a CI pipeline tool.
It shows the status of user configured pipelines and branches for those pipelines.

It looks like this:

image

using config:

```yaml
    buildkite:
      enabled: true
      organizationSlug: "organization-slug"
      refreshInterval: 60
      position:
        top: 2
        left: 0
        height: 1
        width: 9
      pipelines:
        pipeline-slug-1:
          branches:
            - "master"
            - "production"
        pipeline-slug-2:
          branches:
            - "master"
            - "branch-i-want-to-monitor"
```

It's also my first contribution (with go too!), so feel free to point out anything not idiomatic.

The requests are a bit slow because of the size of data, so I will probably switch to the graphql api in the future.

I'll start preparing a PR for the docs as well.

### DIFF
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -6,6 +6,7 @@ import (
 	"github.com/wtfutil/wtf/modules/azuredevops"
 	"github.com/wtfutil/wtf/modules/bamboohr"
 	"github.com/wtfutil/wtf/modules/bargraph"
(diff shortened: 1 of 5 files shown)


senorprogrammer:
Thanks!

Issue #659, opened 2019-09-25
Spotify: support theming


kusold:
## What should it do?
Currently there is no way to re-theme Spotify. It would be nice if this was configurable.
https://github.com/wtfutil/wtf/blob/f5e80119064b9f45e0703da31bfecbbeab9b0b1e/modules/spotify/widget.go#L67-L70

## Screenshot
![image](https://user-images.githubusercontent.com/509966/65647015-6309ab80-dfba-11e9-9f23-3ef3705e6d2b.png)

senorprogrammer:
Added by https://github.com/wtfutil/wtf/pull/677

PR #677, opened 2019-10-04
Support theming for Spotify widget

Tardog:
Solves #659. The changes made here allow users to set custom colors for labels and normal text in the widget.


senorprogrammer:
Excellent, thanks!

PR #679, opened 2019-10-07
Sort Buildkite pipelines and branches alphabetically

jmks:
I noticed that the order of my Pipelines would shuffle between refreshes.
The display data was just looping over a map, which is not ordered in Go.

I introduced a struct to hold the data and display order. I followed the docs on how to sort the `[]Build`s. Since it required introducing a few new types/funcs, I pulled all the display logic to its own type and file.


senorprogrammer:
👍

PR #680, opened 2019-10-07
Make the cmdrunner more interactive

noxer:
Currently the cmdrunner executes the command once for each update interval which delays the update for other widgets and prevents users from using longer running commands (like watching a log or requesting data from very slow servers). This PR fixes the problem by running the command concurrently and updating the view each time the command outputs data.

It also adds two config parameters:
- `tail` which causes the widget to scroll to the end of new output (interesting for following log files)
- `maxLines` which limits the number of lines kept in the buffer (to prevent the widget from consuming unlimited memory)

Also the widget is now focusable to allow the user to scroll in the output.

### DIFF
--- a/modules/cmdrunner/settings.go
+++ b/modules/cmdrunner/settings.go
@@ -7,24 +7,30 @@ import (
 )
 
 const (
-	defaultFocusable = false
+	defaultFocusable = true
 	defaultTitle     = "CmdRunner"
 )
 
+// Settings for the cmdrunner widget
 type Settings struct {
 	common *cfg.Common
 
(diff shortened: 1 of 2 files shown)


noxer:
I've tested the new runner with `dmesg` on Linux.
```yml
    dmesg:
      args: ["--color=always", "-w"]
      cmd: "dmesg"
      tail: true
      enabled: true
      maxLines: 100
      position:
        top: 0
        left: 1
        height: 4
        width: 2
      refreshInterval: 30
      type: cmdrunner
```

senorprogrammer:
Looks good!

PR #681, opened 2019-10-08
added modules cmdrunner2 and resourceusage2

madepolli:
The problem of module cmdrunner is that it can only be used once on the dashboard. I propose a simple extension (could probably be made even simpler, more elegant and with more code reuse) that allows creating cmdrunner2, cmdrunner3, ... and thus multiple cmdrunners.

Second extension (really sorry for making 2 per pull request) is adding an option to not display individual cores in resourceusage plugin. I am working on AMD threadripper, and stats for 32 individual cores are really not interesting to look at. I cloned resourceusage to resourceusage2 and simply changed individual core stats to overall CPU stats. 

The presented solution allows me to stay largely independent of main repo code changes (only widget_maker has to be merged) but I believe the proposed changes could be integrated in a nicer way.

### DIFF
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -8,6 +8,7 @@ import (
 	"github.com/wtfutil/wtf/modules/circleci"
 	"github.com/wtfutil/wtf/modules/clocks"
 	"github.com/wtfutil/wtf/modules/cmdrunner"
+	"github.com/wtfutil/wtf/modules/cmdrunner2"
 	"github.com/wtfutil/wtf/modules/cryptoexchanges/bittrex"
 	"github.com/wtfutil/wtf/modules/cryptoexchanges/blockfolio"
 	"github.com/wtfutil/wtf/modules/cryptoexchanges/cryptolive"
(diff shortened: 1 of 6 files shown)


noxer:
You can run multiple cmdrunners:
```yml
    exe1:
      args: ["args"]
      cmd: "exe"
      enabled: true
      position:
        top: 2
        left: 4
        height: 6
        width: 1
      refreshInterval: 60
      type: cmdrunner
      title: exe1 Title

    exe2:
      args: ["other"]
      cmd: "exe2"
      tail: true
      maxLines: 10000
      enabled: true
      position:
        top: 0
        left: 1
        height: 8
        width: 3
      refreshInterval: 60
      type: cmdrunner
      title: My Other Exe
```

senorprogrammer:
I think the cores issue would be nicer as a configurable option on `ResourceUsage`.

madepolli:
Thanks for the multiple cmdrunners example. I did not know it could be done that way. I went through the documentation and I don't remember seeing such config (it was quite long time ago though).

The resourceusage would be better with an option for overall stats/per core stats, agreed. So I guess this whole pull request is moot.

senorprogrammer:
If you want to do one for resourceusage, that would be awesome.

PR #683, opened 2019-10-09
Expose the widget dimensions in cmdrunner

noxer:
This PR contains two changes:
- Added the utils.CalculateDimensions function to calculate the widget dimensions
- The cmdrunner widget now sets the WTF_WIDGET_WIDTH and WTF_WIDGET_HEIGHT environment variables to allow called commands/scripts to fill the widget

### DIFF
--- a/modules/cmdrunner/settings.go
+++ b/modules/cmdrunner/settings.go
@@ -19,6 +19,10 @@ type Settings struct {
 	cmd      string   `help:"The terminal command to be run, withouth the arguments. Ie: ping, whoami, curl."`
 	tail     bool     `help:"Automatically scroll to the end of the command output."`
 	maxLines int      `help:"Maximum number of lines kept in the buffer."`
+
+	// The dimensions of the module
+	width  int
+	height int
 }
 
 // NewSettingsFromYAML loads the cmdrunner portion of the WTF config
@@ -32,6 +36,8 @@ func NewSettingsFromYAML(name string, moduleConfig *config.Config, globalConfig 
 		tail:     moduleConfig.UBool("tail"),
 		maxLines: moduleConfig.UInt("maxLines", 256),
 	}
+
+	settings.width, settings.height = utils.CalculateDimensions(moduleConfig, globalConfig)
 
 	return &settings
 }
--- a/modules/cmdrunner/widget.go
+++ b/modules/cmdrunner/widget.go
@@ -3,6 +3,7 @@ package cmdrunner
 import (
 	"bytes"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 	"sync"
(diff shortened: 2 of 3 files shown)


senorprogrammer:
[tview](https://www.google.com/search?client=safari&rls=en&q=golang+tview&ie=UTF-8&oe=UTF-8), the underlying library, has a [GetInnerRect](https://godoc.org/github.com/rivo/tview#Box.GetInnerRect) function. Does that do what you're looking for?

And a question about setting the `WTF_WIDGET_WIDTH` and `WTF_WIDGET_HEIGHT` env vars - if there are more than one instance of `CmdRunner` running, will they not conflict?

noxer:
> tview, the underlying library, has a GetInnerRect function. Does that do what you're looking for?

Do I have access to the tview.Box for my widget somehow? It that is the case it would be much more elegant.

> And a question about setting the WTF_WIDGET_WIDTH and WTF_WIDGET_HEIGHT env vars - if there are more than one instance of CmdRunner running, will they not conflict?

The environment variables are set just for the one sub-process, not globally. Multiple instances can not interfere with each other.
I'd also suggest adding an `env` config so users can specify additional variables for their commands.

modules/cmdrunner/settings.go
```diff
@@ -33,5 +37,7 @@ func NewSettingsFromYAML(name string, moduleConfig *config.Config, globalConfig
 		maxLines: moduleConfig.UInt("maxLines", 256),
 	}
 
+	settings.width, settings.height = utils.CalculateDimensions(moduleConfig, globalConfig)
```
Seanstoppable: If we calculate these on start, is it accounted for if we resize the window, or does that not matter?

noxer: AFAIK the tool does not currently support resizing. Please correct me if I'm wrong.

noxer:
> Very interesting. Do you have an example script that leverages these?

I've got a Go program that uses the width to display a departure timetable for my train station:
image

It uses the width data to right-align the times.

noxer:
> tview, the underlying library, has a GetInnerRect function. Does that do what you're looking for?

Oh, the TextView contains View which contains the method. I'll try it, thanks.

Seanstoppable:
@noxer neat!

modules/cmdrunner/settings.go
senorprogrammer: Widget dimensions can be modified in the `config.yml` file while the app is running, and it'll resize the affected widget accordingly.

senorprogrammer:
Looks good, thanks!

modules/cmdrunner/settings.go
noxer: From what I read in the code basically the whole application is restarted. Can a widget somehow detect this?

PR #685, opened 2019-10-10
Fix Errors in azuredevops and devto modules

Boot-Error:
azuredevops had a error due to cli being a pointer to interface rather
than interface.

devto module's `widget.articles` and `displayArticles` data type changed
to `devto.ListedArticle`.

I have fixed the compilation issue I posted on slack channel @senorprogrammer


senorprogrammer:
Awesome!

PR #686, opened 2019-10-10
Rename token to apiKey

jmks:
While writing the docs, I realized there was a standard to call the field `apiKey` in the config, rather than `token`.


senorprogrammer:
Looks good 👍

PR #692, opened 2019-10-11
Secrets file for API Keys

### LINKED ISSUE #517: Security: open call for thoughts on securing WTF's config file
WTF's `config.yml` file is a plain text file stored in the user's home directory. Given the nature of many of the modules in WTF, that file could hold account information, passwords, URIs, and API keys for the user's critical systems. 

This issue is a place to discuss the security of that file.

1. Does it need to be secured? Should we just trust the machine instead?
2. If it should, how so? What's least-friction for the user?

alexfornuto:
This is still very much a work in progress. I'm attempting to help resolve #517 by creating a single file to store API keys (as discussed [here](https://github.com/wtfutil/wtf/issues/517#issuecomment-526019827)), which can then be excluded from config backup and management systems like RCM and the like. 

A solution like this seems better to me than loading these keys as environment variables in the shell, where they could be read in memory by other programs.

I've made the default file and added it to the config, and now my build tests return:

```
app/wtf_app.go:144:5: secrets declared and not used
```

This is where my limited knowledge of Go ends. What I do not yet know how to do is take the contents of the secrets file and load them as environment variables. Anyone who wants to contribute to this branch is welcome to do so. Likewise, any documentation of tutorials you think would be useful in finishing this would be much appreciated.

#hacktoberfest

### DIFF
--- a/app/wtf_app.go
+++ b/app/wtf_app.go
@@ -17,23 +17,25 @@ import (
 // WtfApp is the container for a collection of widgets that are all constructed from a single
 // configuration file and displayed together
 type WtfApp struct {
-	app            *tview.Application
-	config         *config.Config
-	configFilePath string
-	display        *Display
-	focusTracker   FocusTracker
-	pages          *tview.Pages
-	validator      *ModuleValidator
-	widgets        []wtf.Wtfable
+	app             *tview.Application
+	config          *config.Config
+	configFilePath  string
+	secretsFilePath string
+	display         *Display
+	focusTracker    FocusTracker
+	pages           *tview.Pages
+	validator       *ModuleValidator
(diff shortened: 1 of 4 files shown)


noxer:
Go doesn't allow you to declare variables and then not use them. That is why you are getting the error.

alexfornuto:
> Go doesn't allow you to declare variables and then not use them. That is why you are getting the error.

Yes, I understand that. What I don't know is where/how to use them. 

In other words, it's not used because I don't know how to turn YAML into variables, nor where I should be doing that.

senorprogrammer:
Closing this out due to inactivity. Feel free to reopen if anything changes.

PR #695, opened 2019-10-11
Optimizations in ./utils/

noxer:
This PR optimizes a few function in the utils package. Specially the `ToInts` and `ToStrs` functions did not pre-allocate slices and produced unnecessary temporary slices while converting.

- In `conversions.go` the slices are preallocated and the map is created with a size hint. Also the type switch now uses the result of the switch rather than type asserting again.
- In `email_addresses.go` the email slice is preallocated.
- ~In `homedir.go` the `Home` function now just calls `os.UserHomeDir()` which does the exact same thing.~
- In `utils.go` the `ExecuteCommand` function directly writes into a buffer instead of having an intermediate `ioutil.ReadAll` call. Also the `ParseJson` function now decodes directly from the stream (but just one object instead of multiple which may replace the previous ones).

### DIFF
--- a/utils/conversions.go
+++ b/utils/conversions.go
@@ -8,7 +8,7 @@ import (
 
 // MapToStrs takes a map of interfaces and returns a map of strings
 func MapToStrs(aMap map[string]interface{}) map[string]string {
-	results := make(map[string]string)
+	results := make(map[string]string, len(aMap))
 
 	for key, val := range aMap {
 		results[key] = val.(string)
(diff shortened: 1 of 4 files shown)


noxer:
This PR would require Go > 1.12 due to the use of `os.UserHomeDir()`. I'll revert that change.

senorprogrammer:
Feel free to put the `UserHomeDir` improvement back in. The latest version just went out using Go 1.13.1, I just haven't yet updated the documents and `go.mod` file.

noxer:
You need to update Travis CI, it still tries to test the PR against Go 1.11.

senorprogrammer:
I love improvements like this.

Issue #593, opened 2019-09-02
openFileUtil ignored


dbarrosop:
## What's broken?

Looking at the source code looks like [OpenFileUtil](https://github.com/wtfutil/wtf/blob/7e022fe132a9324e712a0fc493f762e4e02e846b/utils/utils.go#L95-L113) is ignored if the target is a URL. That's misleading as it doesn't seem to be documented anywhere. It is also quite limiting as you may want to do something else other than just calling "open". For instance, I am interested in copying the url to the clipboard instead.

It'd be interesting if a `openURL` option could be added (to avoid changing current behavior) and if both `openURLUtil` and `openFileUtil` could be a list instead to support cases where you need to pass options to the util.

Seanstoppable:
Definately not the greatest name. Most of the use cases are actually opening URLs, with one use case to open a text file. The code is effectively trying to hook into the system handler for handling a URI, which I don't think is wrong and probably actually works for the general file system case.

I do think the copying of the url to the clipboard concept is interesting and would like to dive into it. Wondering what use case here is for copying a url to the clipboard, but NOT opening the browser.

Also wondering in what case we want to support a list, since all selection is currently done on a single item afaikt?

dbarrosop:
My use case is quite straightforward, not sure how common though. My laptop is a macbook but I do all my coding and most of my work in a VM, which means I am also running wtf inside that VM. As I have the VM and the macos clibpboards connected via a socket sending the URL to the linux clipboard means I can easily grab it and paste it in the browser running on macos. On the tmux side I use `DanielFGray/tmux-fzf-urls` for that very same purpose.

dbarrosop:
Yes, I have the following script:

```
#!/bin/bash
PORT=$1
TEXT=${@:2}
nc -N localhost $PORT <<< $TEXT
```

So in the shell I can just do:

```
$ copy_to_clipboard.sh 9999 this will end up in my macos clipboard
```

Port 9999 is tunneled to my macOS clipboard so whatever is sent there ends up in my clipboard

PR #698, opened 2019-10-12
Add new urlutil option, and have these be reloadable

Seanstoppable:
Handles #897 and part of #593


senorprogrammer:
Looks good 👍

PR #706, opened 2019-10-19
Improve gitlab calls

### LINKED ISSUE #703: Improvement: selectively fetch GitLab resources
The current implementation fetches **all** MRs and Issues belonging to a project, then loops through them to find relevant items to be displayed.

For large projects, this creates a lot of data transfer and also takes time to loop.

It could fetch specifically authored or assigned resources, and possibly perform them concurrently.

caalberts:
A few changes introduced in this PR:
1. Adding user context to GitLab project, which allows us to fetch issues and MRs selectively by the user, reducing the need to filter for the user when displaying data.
2. Renaming Approval Requests to Assigned Merge Requests, as there is actually no "approvals" in a MergeRequest until it has been approved.

The call to get all open Merge Requests is still present because it is needed to count open Merge Requests.

Fixes #703

### DIFF
--- a/modules/gitlab/display.go
+++ b/modules/gitlab/display.go
@@ -22,8 +22,8 @@ func (widget *Widget) content() (string, string, bool) {
 	str += " [red]Stats[white]\n"
 	str += widget.displayStats(project)
 	str += "\n"
-	str += " [red]Open Approval Requests[white]\n"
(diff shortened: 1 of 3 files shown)


senorprogrammer:
Thanks!

Issue #657, opened 2019-09-24
Improve test suite coverage


senorprogrammer:
## What should it do?

Now that Dependabot is in place and monitoring outdated dependencies, it would be great to have it also auto-merge those kinds of PRs. For that to be reasonable, WTF needs much better test suite coverage.

Improve test suite coverage.

Tardog:
I would like to work on expanding the test suite. Does this entail writing tests for all of the modules, too, or just for the core app? What amount of coverage would be sufficient? Do we want to cover as many execution paths as possible, or is there merit in focusing on critical parts of the code?

senorprogrammer:
Right now the coverage is so sparse that adding anything anywhere would be a great improvement. Areas that would have the most impact to start would be the ones shared amongst all the modules: `/app`, `/cfg`, `/view`, and `/wtf`.

Tardog:
Got it. I'm going to focus on these areas first.

PR #710, opened 2019-10-21
Add tests for bargraph view

hemu:
Add tests for all exported methods for BarGraph view.

To chip away at #657. If this PR is in the right direction I can work on adding additional tests for other pkgs to improve coverage.


senorprogrammer:
Great, thanks!

PR #715, opened 2019-10-22
Twitterstats

Ameobea:
This PR implements a new `twitterstats` module which displays the number of followers and tweets for a list of users:

I created it for personal use, and I totally understand if you don't think it would be a good fit for merging into the official project.  That being said, if you would like this to be a part of official wtfutil, I'd be happy to create a docs entry for it and tweak it/add features as required!

### DIFF
--- a/modules/twitter/client.go
+++ b/modules/twitter/client.go
@@ -1,9 +1,14 @@
 package twitter
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"strconv"
+
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/clientcredentials"
 )
 
 /* NOTE: Currently single application ONLY
@@ -12,19 +17,37 @@ import (
 
 // Client represents the data required to connect to the Twitter API
 type Client struct {
-	apiBase     string
-	bearerToken string
-	count       int
-	screenName  string
+	apiBase    string
+	count      int
+	screenName string
+	httpClient *http.Client
 }
 
 // NewClient creates and returns a new Twitter client
 func NewClient(settings *Settings) *Client {
+	var httpClient *http.Client
+	// If a bearer token is supplied, use that directly.  Otherwise, let the Oauth client fetch a token
+	// using the consumer key and secret.
+	if settings.bearerToken == "" {
+		conf := &clientcredentials.Config{
+			ClientID:     settings.consumerKey,
+			ClientSecret: settings.consumerSecret,
+			TokenURL:     "https://api.twitter.com/oauth2/token",
+		}
+		httpClient = conf.Client(oauth2.NoContext)
+	} else {
+		ctx := context.Background()
+		httpClient = oauth2.NewClient(ctx, oauth2.StaticTokenSource(&oauth2.Token{
+			AccessToken: settings.bearerToken,
+			TokenType:   "Bearer",
+		}))
+	}
+
 	client := Client{
-		apiBase:     "https://api.twitter.com/1.1/",
-		count:       settings.count,
-		screenName:  "",
-		bearerToken: settings.bearerToken,
+		apiBase:    "https://api.twitter.com/1.1/",
+		count:      settings.count,
+		screenName: "",
+		httpClient: httpClient,
 	}
 
 	return &client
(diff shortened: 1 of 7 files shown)


modules/twitterstats/client.go
```diff
@@ -0,0 +1,95 @@
...
+	usernames := make([]string, len(settings.screenNames))
+	for i, username := range settings.screenNames {
+		switch username.(type) {
+		default:
+			{
+				log.Fatalf("All `screenName`s in twitterstats config must be of type string")
+			}
+		case string:
+			usernames[i] = username.(string)
+		}
```
noxer: This whole switch can be represented by
```go
if usernames[i], ok = username.(string); !ok {
    log.Fatalf("All `screenName`s in twitterstats config must be of type string")
}
```
(you should declare `var ok bool` before the loop)

modules/twitterstats/client.go
```diff
@@ -0,0 +1,95 @@
...
+		stats[i] = TwitterStats{
+			followerCount: 0,
+			tweetCount:    0,
+		}
+
+		res, err := client.httpClient.Get(fmt.Sprintf("%s?screen_name=%s", userTimelineURL, username))
+		if err != nil {
+			continue
+		}
+		defer res.Body.Close()
```
noxer: This will close the Body when the whole function terminates, which will prevent `http` from reusing the connection for other requests in this loop. You should consider to factor the request + decode into its own function and just call it here.

modules/twitterstats/client.go
```diff
@@ -0,0 +1,95 @@
...
+
+		body, err := ioutil.ReadAll(res.Body)
+		if err != nil {
+			continue
+		}
+
+		var parsed map[string]interface{}
```
noxer: When you declare `TwitterStats` as
```go
type TwitterStats struct {
	FollowerCount int64 `json:"followers_count"`
	TweetCount    int64 `json:"statuses_count"`
}
```
you can unmarshal directly into a `TwitterStats` struct.

Ameobea:
@noxer tyvm for the thoughtful review; go is not a strong language of mine and I appreciate the info.  I'll make those changes asap!

modules/twitterstats/client.go
Ameobea: That's a really neat functionality - very cool that's built directly into go.

senorprogrammer:
This is looking good. Could you include an example of the configuration for this module? Without an example it's difficult to test and write docs for.

Also, two questions: 
* where do `clientID` and `clientSecret` come from? Could you provide links to the Twitter docs for those
* the `twitter` module uses the same API as this one (`https://api.twitter.com/1.1/`) and uses a `bearerToken` in the header to do authentication. Is it possible for these two Twitter-related modules to share a common authentication? That would make things simpler for users

senorprogrammer:
I would love to get this in before [Hacktoberfest](https://hacktoberfest.digitalocean.com) ends. If you think you can get any changes done before the 31st, let me know and I'll wait. If not, let me know and I'll merge in as-is and we can use another PR for anything else.

Ameobea:
@senorprogrammer I'll finish this up tonight or tomorrow including getting the auth matched between this and the other twitter module and writing up docs + example config.

noxer:
@senorprogrammer having merged the PR is no requirement for the Hacktoberfest challenge. It can even be closed without a merge. Just marking the PR as invalid excludes it from being counted.

Ameobea:
Both auth methods (bearer token and consumer API keys) are now supported by both Twitter modules.

I've created docs for the twitterstats module and updated docs for the twitter tweets module to reflect the ability to use both auth schemes: https://github.com/wtfutil/wtfdocs/pull/78

Please let me know if there's any other changes you'd like made.

modules/twitter/client.go
```diff
@@ -12,19 +17,37 @@ import (
...
-	screenName  string
+	apiBase    string
+	count      int
+	screenName string
+	httpClient *http.Client
 }
 
 // NewClient creates and returns a new Twitter client
 func NewClient(settings *Settings) *Client {
+	var httpClient *http.Client
```
Ameobea: These several lines are duplicated verbatim between this file and `modules/twitterstats/client.go`.  I don't know where to put them so that they can be shared, but if anyone has suggestions I'll do it.

PR #720, opened 2019-10-23
Adds football fixtures/standings widget

C123R:
#hacktoberfest

Hi there,

If this merges, it will add a widget for football fixtures/standings. In order to use this widget, you need to register to [football-data](https://www.football-data.org/client/register) to get api key(Free tier).

image

config.yaml:
```yaml

wtf:
  colors:
    border:
      focusable: darkslateblue
      focused: orange
      normal: gray
  grid:
    columns: [32, 32, 32, 32, 90]
    rows: [10, 10, 10, 4, 4, 90]
  refreshInterval: 1
  mods:
    football:
      enabled: true
      apiKey: "*******"              # football-data.org api key
      league: "SPD"                   # shortname for the league
      favTeam: "FC Barcelona"        # Favorite team from specific league, Default: ""
      standingCount: 10                  # Top teams count, Default: 5
      matchesFrom: 5           # Matches till Today (Today - Number of days), Default: "2"
      matchesTo: 7                  # Matches from Today (Today + Number of days), Default: "5"
      position:
        top: 1
        left: 0
        height: 3
        width: 3
      refreshInterval: 300
      title: "⚽"

Please let me know if there is anything which can be improved :)

I will create PR for the documentation as well if all looks good here.

### DIFF
--- a/modules/football/widget.go
+++ b/modules/football/widget.go
@@ -153,17 +153,19 @@ func (widget *Widget) GetMatches(leagueId int) string {
 
 	if len(l.Matches) != 0 {
 
-		for _, val := range l.Matches {
-			if strings.Contains(val.AwayTeam.Name, widget.settings.favTeam) || strings.Contains(val.HomeTeam.Name, widget.settings.favTeam) || widget.settings.favTeam == "" {
-				if val.Status == "SCHEDULED" {
-					row := []string{"⚽", val.HomeTeam.Name, "🆚", val.AwayTeam.Name, parseDateString(val.Date)}
-					tScheduled.Append(row)
-				} else if val.Status == "FINISHED" {
-					row := []string{"⚽", val.HomeTeam.Name, strconv.Itoa(val.Score.FullTime.HomeTeam), "🆚", val.AwayTeam.Name, strconv.Itoa(val.Score.FullTime.AwayTeam)}
-					tPlayed.Append(row)
-				}
+		for _, m := range l.Matches {
+
+			widget.markFavorite(&m)
+
+			if m.Status == "SCHEDULED" {
+				row := []string{m.HomeTeam.Name, "🆚", m.AwayTeam.Name, parseDateString(m.Date)}
+				tScheduled.Append(row)
+			} else if m.Status == "FINISHED" {
+				row := []string{m.HomeTeam.Name, strconv.Itoa(m.Score.FullTime.HomeTeam), "🆚", m.AwayTeam.Name, strconv.Itoa(m.Score.FullTime.AwayTeam)}
+				tPlayed.Append(row)
 			}
 		}
+
 		tScheduled.Render()
 		tPlayed.Render()
 		if playedBuf.String() != "" {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by massa1240** [comment]
i would squash both commits `support keyboard` to make them only

**Comment by massa1240** [comment]
Some feature feedback:
what is the behaviour of `favTeam`? I can see how this can be extended in two ways, but I didn't see if these already happen:
(1) I would like to follow the upcoming games, but want to highlight my fav team
(2) I want to hide all matches and only display my favourite(s) team.
As a feature, it looks nice having options for both behaviour :) 

I also believe that favourite team and leagues could be a list! Some people are really crazy about following multiple leagues haha Is it already possible?

**Comment by C123R** [comment]
> i would squash both commits `support keyboard` to make them only

Done 👍

File: modules/football/widget.go [hunk 1]
```diff
@@ -0,0 +1,181 @@
...
+	}
+	err = json.Unmarshal(data, &l)
+	if err != nil {
+		return fmt.Sprintf("Error fetching matches: %s", err.Error())
+	}
+
+	if len(l.Matches) != 0 {
+
+		for _, val := range l.Matches {
+			if strings.Contains(val.AwayTeam.Name, widget.settings.favTeam) || strings.Contains(val.HomeTeam.Name, widget.settings.favTeam) || widget.settings.favTeam == "" {
```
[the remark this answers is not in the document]
C123R: Sure, that was kind of early version which I wanted to fix later on..but missed :)

As suggested, I have extracted it as a function that will mark the fav team with Star. Also, it will print reset other fixtures or past matches.

image

modules/football/widget.go
```diff
@@ -0,0 +1,181 @@
...
+	"EL2": {444, "Campeonato Brasileiro da Série A"},
+	"CL":  {2001, "UEFA Champions League"},
+	"FL1": {2015, "French Ligue 1"},
+	"GB":  {2002, "German Bundesliga"},
+	"ISA": {2019, "Italy Serie A"},
+	"NE":  {2003, "Netherlands Eredivisie"},
+	"PPL": {2017, "Portugal Primeira Liga"},
+	"SPD": {2014, "Spain Primera Division"},
+	"WC":  {2000, "FIFA World Cup"},
+}
```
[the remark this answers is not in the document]
C123R: Agree. This is for sure in my todo but will fix it as an issue in a further release. For now, I will mention the list of leagues that are supported in Documentation. 

I will come up with another PR for the doc as soon as this PR is merged.

C123R:
> Some feature feedback:
> what is the behaviour of `favTeam`? I can see how this can be extended in two ways, but I didn't see if these already happen:
> (1) I would like to follow the upcoming games, but want to highlight my fav team

I guess this is already covered in this PR.

> (2) I want to hide all matches and only display my favourite(s) team.

Sure.

> As a feature, it looks nice having options for both behaviour :)
> 
> I also believe that favourite team and leagues could be a list! Some people are really crazy about following multiple leagues haha Is it already possible?

Currently not, but totally make sense of having this option.

Thanks for the feedback :) Will take care of this feature in separate PR. 👍

modules/football/widget.go
```diff
@@ -0,0 +1,196 @@
...
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Sprintf("Error fetching standings: %s", err.Error())
+	}
+	err = json.Unmarshal(data, &l)
+	if err != nil {
+		return fmt.Sprintf("Error fetching standings")
+	}
+
+	if len(l.Standings) > 0 {
```
noxer: If you flip this condition, you can avoid the indentation of the loop.

modules/football/widget.go
```diff
@@ -0,0 +1,196 @@
...
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return fmt.Sprintf("Error fetching matches: %s", err.Error())
+	}
+	err = json.Unmarshal(data, &l)
+	if err != nil {
+		return fmt.Sprintf("Error fetching matches: %s", err.Error())
+	}
+
+	if len(l.Matches) != 0 {
```
noxer: Flip this condition and return early to avoid one indentation for the following code.

C123R: Thanks for the hint, fixed in latest commit :)

senorprogrammer:
This is very cool, thanks for adding it!

PR #724, opened 2019-10-25
modded resourceusage

madepolli:
Modified resourceusage to allow for combined cpu usage (useful on machines with lots of cores and threads)

Change includes the config (see below). The variable `cpuCombined bool` is new - default is false, which makes the default behaviour equal to the old behaviour, i.e., not combined stats. Setting it to true, combines the core stats into one, and changes its title to 'CPU'.

There is also a change in widget Refresh function, which makes it flicker-free for me (it would flicker before, but I don't know the reasoning for why was it setup as it was before, so it might break something else).

Example of config:
```yaml
    resourceusage:
      enabled: true
      position:
        top: 0 
        left: 0
        height: 1
        width: 1
      refreshInterval: 1
      cpuCombined: true
```

### DIFF
--- a/modules/resourceusage/settings.go
+++ b/modules/resourceusage/settings.go
@@ -12,11 +12,13 @@ const (
 
 type Settings struct {
 	common *cfg.Common
+	cpuCombined bool
 }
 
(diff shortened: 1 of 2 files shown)


senorprogrammer:
This is great, thanks

PR #743, opened 2019-11-10
Update goproxy to default proxy.golang.org

chenrui333:
Recently, when I tried to update `go.mod` file in goreleaser project, we found that gocenter hash for some dependency is not consistent with proxy.golang.org (which is the default goproxy).

Details are in this comment, https://github.com/goreleaser/goreleaser/pull/1217#issuecomment-547150158
And the golang proxy updated in this PR, https://github.com/goreleaser/goreleaser/pull/1220#issuecomment-547639705

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,6 +1,6 @@
 env:
   - GO111MODULE=on
-  - GOPROXY="https://gocenter.io"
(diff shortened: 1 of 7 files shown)


senorprogrammer:
I like this. Thanks!

PR #746, opened 2019-11-12
Added a setting to hide the completed downloads in the transmission module

schoentoon:
This adds a setting to the existing transmission module called `hideComplete` which when set to true will simply hide all the torrents that are at 100% from the view.


senorprogrammer:
Sweet!

PR #749, opened 2019-11-12
Exchangerates module

schoentoon:
This adds a new module to show exchange rates, it fetches these from https://exchangeratesapi.io/
Example configuration below

```
exchangerates:
      rates:
        EUR: "CHF"
        CHF:
          - "USD"
          - "EUR"
```

In this case it'll show the exchange rate from EUR to CHF and from CHF to USD and EUR.


senorprogrammer:
Nice!

schoentoon:
Sadly I found a bit of an annoying bug with it just yesterday, followup pull request will follow.

PR #752, opened 2019-11-16
Bugfix for exchangerates module

schoentoon:
As I indicated in #749 I found a tiny bug that should be fixed with this release. If you're running the exchangerates module with a refreshInterval it could be that throughout the runtime of wtfutil the order that results would appear in change randomly. This merge request simply keeps the order and displays it accordingly.


senorprogrammer:
Nice catch. Will merge today

Issue #753, opened 2019-11-16
Subreddit-When top links is fewer than widget.settings.numberOfPosts


TDHTTTT:
## What's broken?

In subreddit module, when top links from the reddit is acutally fewer than widet.settings.numberOfPosts, it will complain `panic: runtime error: slice bounds out of range`

TDHTTTT:
I tried to fix it in #754

senorprogrammer:
Fixed by https://github.com/wtfutil/wtf/pull/754

PR #754, opened 2019-11-17
subreddit link bound checking

TDHTTTT:
When subreddit link length is less than widet.settings.numberOfPosts, widget.links is set to be links.


TDHTTTT:
As reported in #753

senorprogrammer:
Thanks!

PR #778, opened 2019-12-08
Prevent flickering in cmdRunner widgets

### LINKED ISSUE #732: CmdRunner problems after improvements in 0.23
## What's broken?

After the improvements of #680 there are two problems:

1. The refresh of CmdRunner instances is visible when ansi color sequences are used, the module is briefly blank before the content is refreshed. 

2. The tail configuration does not work, at least with 'tail: false', the output still scrolls. I wanted to use it to get the old behaviour back.

Also the new configuration options are not in the documentation on the wtfutil.com.

Gibstick:
This commit removes flickering in the cmdRunner widgets while preserving the live-update functionality. It amends https://github.com/wtfutil/wtf/commit/45b955 by not redrawing on every write call. Instead, the logic in Refresh is as follows:

1. If the command is already running, it will not try to re-run the
command. The default case in the select will trigger a re-draw instead
so that new output can be seen. This accommodates long-runing commands
eg. tailing a log.
2. If the command is not already running, it will trigger a new run.
When the command terminates, it will trigger a re-draw. This means the
widget refreshes as soon as possible, to accommodate the original use
case of running a command and displaying its output in the widget.

In all cases, the widget will not re-draw more often than the refresh interval. This is what eliminates flickering, since the previous implementation before using goroutines was not redrawing more than once per refresh interval.

I haven't written a lot of Go code before so don't hesitate to nitpick the style etc.

Partially fixes #732 (I don't know what's wrong with the tail configuration so I didn't try to fix it here).

### DIFF
--- a/modules/cmdrunner/widget.go
+++ b/modules/cmdrunner/widget.go
@@ -18,9 +18,10 @@ type Widget struct {
 
 	settings *Settings
 
-	m       sync.Mutex
-	buffer  *bytes.Buffer
-	running bool
+	m          sync.Mutex
+	buffer     *bytes.Buffer
+	runChan    chan bool
+	redrawChan chan bool
 }
 
 // NewWidget creates a new instance of the widget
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Awesome work, thanks!

PR #779, opened 2019-12-09
Enable Jenkins module to work with Multi-configuration projects

NickyMateev:
# Enable Jenkins module to work with Multi-configuration projects

In practice it's quite often for teams to have Jenkins configurations not in just a plain Jenkins Project/View, but to rather have numerous Multi-configuration projects.

For such cases it would be very useful if the Jenkins module for `wtf` could support Multi-configuration projects.

This change aims to enable this scenario and thus further evolve the Jenkins module.

> Note: I've removed some configuration properties that were never used (and were actually somewhat incorrectly defined).

### DIFF
--- a/modules/jenkins/client.go
+++ b/modules/jenkins/client.go
@@ -45,10 +45,13 @@ func (widget *Widget) Create(jenkinsURL string, username string, apiKey string) 
 		return view, err
 	}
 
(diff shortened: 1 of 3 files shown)


senorprogrammer:
Awesome, thanks!

PR #786, opened 2019-12-14
WTF-748 Fixes and improves module templating

### LINKED ISSUE #748: Module template seems quite outdated
## What's broken?
After having followed the instructions at https://wtfutil.com/configuration/generate-module/ and started working on my module, I first had to figure out that I had to add my widget to the long list at https://github.com/wtfutil/wtf/blob/master/app/widget_maker.go
After which I came to the conclusion that a newly generated module wouldn't compile. The `wtf.TextWidget` had to be changed to `view.TextWidget` for example.
Perhaps this template should be updated, maybe the generate part could output that you should add your widget to https://github.com/wtfutil/wtf/blob/master/app/widget_maker.go
Perhaps it could even be added to integration tests so this breaking in the future will sooner be realized.

senorprogrammer:
Fixes and improves the module templating for creating new bare-bones
text widgets.

This command:

    WTF_WIDGET_NAME=MyNewWidget go generate -run=text

now properly generates:

    * the module directory in the /modules directory
    * the widget.go file
    * the settings.go file

with no linter warnings or errors.

Fixes #748 

Signed-off-by: Chris Cummer <chriscummer@me.com>

### DIFF
--- a/generator/settings.tpl
+++ b/generator/settings.tpl
@@ -1,0 +1,29 @@
+package {{(Lower .Name)}}
+
+import (
+	"github.com/olebedev/config"
+	"github.com/wtfutil/wtf/cfg"
+)
+
+const (
+	defaultFocusable = false
+	defaultTitle     = "{{(.Name)}}"
+)
+
+// Settings defines the configuration properties for this module
+type Settings struct {
+	common *cfg.Common
+
+    // Define your settings attributes here
+}
+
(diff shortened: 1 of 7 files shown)


schoentoon:
Would it be an idea to execute this from travis CI so we'll be certain this remains up to date? Could just be an easy `go generate`  followed by a `go build`.

PR #808, opened 2020-01-07
Added SpaceX Module

bjoernw:
Next SpaceX Launch Widget 

No config parameters since the API is unauthenticated [API Docs](https://docs.spacexdata.com/?version=latest#c75a20cf-50e7-4a4a-8856-ee729e0d3868)

Screen Shot 2020-01-06 at 11 23 54 PM

If widget height >= 2 it also displays a description of the launch. It can be quite a bit of text, hence the height requirement.

I'm a big fan of watching these live so this will be a nice reminder.


senorprogrammer:

PR #812, opened 2020-01-09
Upgrade tview

### LINKED ISSUE #618: Colors won't work for cmdrunner when using nodejs chalk or colorette library
## What's broken?
Colors won't work for cmdrunner when using nodejs chalk or colorette library

Seanstoppable:
Should fix #618


senorprogrammer:

PR #813, opened 2020-01-09
Support XDG_CONFIG_HOME

### LINKED ISSUE #699: XDG_CONFIG_HOME Should Be Used When Available
The environment variable `XDG_CONFIG_HOME` should be used when available as mentioned in the Free Desktop specification: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html

I see close mentions of this in the source code right now but it's not actually implemented.

I can work on this if it's agreed that this should be worked on.

Seanstoppable:
Clean up workflow and eliminate a method
One place where we calculate the full dir

Closes #699


senorprogrammer:

PR #817, opened 2020-01-12
Dockerfile - Use progressive build to reduce final image size

firecat53:
- Use progressive build to reduce final image size
- Download source during build so only Dockerfile is required to be present on the build host
- Add a `version` arg so any tag or branch can be built: `docker build --build-arg=version=v0.25.0 -t wtf .`


senorprogrammer:

PR #856, opened 2020-03-05
Support to enable(default)/disable display of CPU, mem and swap

leterio:
Add options to enable or disable display CPU, mem and swap information on resource usage widget


senorprogrammer:
Thanks. I'll try to merge this in this weekend.

PR #874, opened 2020-04-13
add pihole module.

jonhadfield:
Hi, this is a new module to display information about a running a [Pi-hole](https://pi-hole.net/) server.  
This doesn't introduce any new package dependencies but calls the Pi-Hole API endpoint using the standard library and displays them using the already included [tablewriter](github.com/olekukonko/tablewriter) library.  
As a first attempt at writing a module, I've based this on existing modules, so not sure I'm following best practices for this project.
Submitting PR for docs shortly.
Any feedback appreciated.

### DIFF
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -46,6 +46,7 @@ import (
 	"github.com/wtfutil/wtf/modules/newrelic"
 	"github.com/wtfutil/wtf/modules/opsgenie"
 	"github.com/wtfutil/wtf/modules/pagerduty"
+	"github.com/wtfutil/wtf/modules/pihole"
 	"github.com/wtfutil/wtf/modules/pocket"
 	"github.com/wtfutil/wtf/modules/power"
 	"github.com/wtfutil/wtf/modules/resourceusage"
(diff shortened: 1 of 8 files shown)


senorprogrammer:
If you run `go mod tidy` on your PR, the CI checks should pass.

senorprogrammer:
I want to use this, but I haven't been able to get past the following error:

```
failed to connect to Pi-hole server: Get "?version=": unsupported protocol scheme ""
```
with an API URL of: `http://192.168.x.xx/admin/api.php`

Any thoughts?

I also recommend having your error messages output the error itself, instead of `url.String()`. I replaced the failing one with 

```
return fmt.Errorf(" failed to connect to Pi-hole server: %v", err.Error())
```
to get the message above.

jonhadfield:
To start with the error itself, in your web client you should be able to do a GET with http://192.168.x.x/admin/api.php?version (no auth token required) to get a json response with the version number.
According to the [API doc](http://192.168.4.50:1010/admin/api.php?version) this should work with both API versions 2 and 3 (TBH, I've not tried against version 2 and just trusted the doc). The spurious `=` after `?version` is an artefact of using url.Values to build query strings, so I might just drop that for this call. Please could you also check you're not missing a port number and are really running on port 80?

I tried to avoid outputting the full error due to the amount of space available and, probably naively, trusting there weren't enough scenarios where it would be useful. I'm also concious of the full error disclosing the API URL and, most importantly, token if used outside of a home environment. What I'll do is output the error but strip the token from any error responses.
I'll also include a message about a potentially missing port number though as it might catch some users out.

jonhadfield:
I've made the changes mentioned above and all errors are now output but auth token is stripped.  
wtf docs PR updated: https://github.com/wtfutil/wtfdocs/pull/87
Rebased against master.

senorprogrammer:
That is working nicely!

senorprogrammer:
Merged. This is an awesome addition.

PR #877, opened 2020-04-17
Add event end time to gcal module display

acaloiaro:
During a busy day packed with meetings, it's important to be aware of the gaps between each meeting. Seeing event end times helps you plan your day better.


senorprogrammer:
I think this is a good idea, and it should be opt-in so as not to break the layouts of existing users. Could you put this behind a config option for gCal that defaults to `false`?

acaloiaro:
Thanks @senorprogrammer -- the PR has been updated to add a setting to show event end times, and I've added a bit of test coverage over the new behavior.

senorprogrammer:
Thanks!

PR #882, opened 2020-04-27
Gitlab: allow for opening items in browser

mogensen:
Implementing more keyboard actions for the GitLab module.
This stream lines both the code and the experience with the Github module.


senorprogrammer:
Very nice!

Issue #811, opened 2020-01-08
Random capital letters around


andersmmg:
## What's broken?
I get random capital letters seemingly randomly strewn across the page. For a while it was capital "A", now they're capital "P". Is this supposed to be here or...
Edit: here's a screen shot, it's just a corner since I have a lot of personal info.
https://imgur.com/d2CNlT8.png

senorprogrammer:
Could you post the OS your running this on?

andersmmg:
OS: Manjaro 18.1.5 Juhraya
Kernel: x86_64 Linux 5.4.6-2-MANJARO

senorprogrammer:
And no, those aren’t supposed to be there. Something is definitely glitched.

Seanstoppable:
@andersmmg can you also share the value of $TERM?

BonoVanderpoorten:
Hey,
Having the same issue, I'm on WSL in the new Windows Terminal though.
I suffer the letter P and they seem to pop up over time, zooming / tabbing may make them appear more frequently. 

Img:
https://i.imgur.com/6alDGR2.png

willjohnson:
I am having the filepath of my textfile widget randomly written to my screen. Possibly related?

GideonWolfe:
+1, Also experiencing this issue on `Arch Linux 5.6.2` with `urxvt`. Random `A`s and `P`s being dropped all over the terminal.

jonhadfield:
Appears to be a bug in the digitalclock module.  
I've submitted a fix in https://github.com/wtfutil/wtf/pull/883.

senorprogrammer:
Fixed by https://github.com/wtfutil/wtf/pull/883

PR #883, opened 2020-04-27
fix digitalclock display bug.

jonhadfield:
This is a fix for https://github.com/wtfutil/wtf/issues/811.
I believe the issue is in the digitalclock module where the developer has accidentally left a print statement.

I was able to recreate with the following config:
```
digitalclock:
      color: orange
      enabled: true
      font: bigfont
      hourFormat: 12
      position:
        top: 0
        left: 0
        height: 1
        width: 1
      refreshInterval: 1
      title: "big clock"
      type: "digitalclock"
```
and you can see the P at the bottom-left and on the right-hand side. 
image
The clue was in Epoch showing a P after the number, when Epoch doesn't include AM/PM. It's not clear that @andersmmg is using the module, but the random A or P (for AM or PM) would fit the behaviour. It should at least fix issue seen by @BonoVanderpoorten with this module.

### DIFF
--- a/modules/digitalclock/clocks.go
+++ b/modules/digitalclock/clocks.go
@@ -40,7 +40,6 @@ func getHourMinute(hourFormat string) string {
 	}
 
 	strMintues := intStrConv(time.Now().Minute())
-	fmt.Println(AMPM)
 	strMintues = strMintues + AMPM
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Thanks!

PR #885, opened 2020-04-30
Add ipv6 support to ipinfo

acaloiaro:


acaloiaro:
This was admittedly a sort of quick and dirty approach to determining one's ostensibly routable ipv6 address. I wonder if it would be more appropriate to simply check `net.Interfaces()` and determine if any are routable using the system's routing table. There may be cross-platform concerns there, though.

acaloiaro:
There's a much cleaner way to detect if one has an internet-routable IPv6 addr using  https://pkg.go.dev/mod/v.io/x/lib.

After having learned a bit more about IPv6, I see that the same ends can be achived without dialing out and relying on a third-party service with a `AAAA` DNS record present. But with that said, this module isn't mission-critical and already depends upon a 3rd party service (ipinfo.io). Relying on Netflix's `fast.com` to determine one's internet-routable IPv6 addr seems a small downside in comparison to pulling in a new project dependency. 

Thoughts?

senorprogrammer:
I like the approach you currently have, because it's relatively simple, given what's already there (dependencies on external connections aside). That said, if you want to go the Vanadium route, I'm all for that as well. I don't mind the dependency coming in.

acaloiaro:
Sure, I'll leave this PR as-is then. I also prefer the simplicity of this approach.

senorprogrammer:
Thanks!

PR #886, opened 2020-04-30
Fix feed module item sorting

rsaarelm:
The original one doesn't sort the items at all, you can see this when you have multiple feeds and the items are never interleaved. Turns out `gofeed` already does date parsing, switch to using its parsed dates and it starts working.


senorprogrammer:
Thanks!

PR #895, opened 2020-05-16
golangci-lint configuration file

sam-github:
golangci-lint can run all the currently enabled linters, and as far as I
can tell, does it in under 5 seconds as opposed to over 180 seconds
(compare `time make cilint` and `time make lint`).

Some of the linters that are listed in the "enabled" section but
commented out looked like a good idea to me, and fairly low hanging
fruit to fix, but they are not passing at the moment.

All the linters covered in the current Makefile are run.

TODO:
- replace lint target in Makefile with golangci-lint
- remove .github/workflow/errcheck.yml

-----

Just a suggestion. It might be much faster, and yet do the same thing, win-win. I did some fixing of things `gocritic` didn't like (its a big fan of switch statements over if/if else/if else/else series), but then stopped and disabled it. Some of the sec warnings seem not applicable to wtf given that it does shell out to user-defined commands at times, so enabling gosec would involve whitelisting some of the warnings. Etc.

On the down side, since this is passing, I won't see the line-annotation feature that is supposed to be enabled... maybe I should enable a couple of the failing linters temporarily in the PR, so that feature is visible?

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,20 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+  pull_request:
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v0.2.0
+        with:
+          # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
+          version: v1.27
+          args: ./...
--- a/Makefile
+++ b/Makefile
@@ -86,6 +86,9 @@ install:
 	@echo "${APP} installed into ${INSTALLPATH}"
 
 ## lint: runs a number of code quality checks against the source code
+cilint:
+	golangci-lint run
+
 lint:
 	@echo "\033[35mhttps://github.com/kisielk/errcheck\033[0m"
 	errcheck ./app
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,15 @@
+linters:
+  enable:
+    - vet
+    - errcheck
+    - staticcheck
+#   - dupl
+#   - funlen
+#   - goconst
+#   - gocritic
+    - gofmt
+#   - golint
+#   - misspell
+#   - stylecheck
+    - unconvert
+#   - whitespace
--- a/app/focus_tracker.go
+++ b/app/focus_tracker.go
@@ -200,7 +200,7 @@ func (tracker *FocusTracker) focusables() []wtf.Wtfable {
 	}
 
 	// Sort for deterministic ordering
-	sort.SliceStable(focusable[:], func(i, j int) bool {
+	sort.SliceStable(focusable, func(i, j int) bool {
 		iTop := focusable[i].CommonSettings().Top
 		jTop := focusable[j].CommonSettings().Top
 
--- a/app/wtf_app.go
+++ b/app/wtf_app.go
@@ -109,6 +109,7 @@ func (wtfApp *WtfApp) keyboardIntercept(event *tcell.EventKey) *tcell.EventKey {
 		switch string(event.Rune()) {
 		case "/":
 			return nil
+		default:
 		}
 	}
 
--- a/cfg/config_files.go
+++ b/cfg/config_files.go
@@ -78,7 +78,7 @@ func WtfConfigDir() (string, error) {
 	if configDir == "" {
 		configDir = WtfConfigDirV2
 	} else {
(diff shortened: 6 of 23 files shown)


sam-github:
Its possible (see https://github.com/golangci/golangci-lint/issues/825) that this is because `go build` (no args) tries to make a `./wtf` executable, but there is already a `./wtf/` directory.

sam-github:
The inline annotations are nice (the existing integration likely has them, too, that's not an advantage).

Turns out it was just a `./...` arg that's needed in the CI action (though not locally).

sam-github:
Times for runs can be seen above. golangci-lint was 1m, imports check 51 sec, the others are 2m+. A small win.

senorprogrammer:
This will be great on the CI system. For local installs we'll need to ensure that the user has it installed. What do you think of changing `make lint` to check that `golangci-lint` is present. If it is, use that. If not, inform on where to get it and use the old method.

senorprogrammer:
And this change to the original `lint` command speeds it up quite a bit, making it only 7x slower than `golangci-lint` (the original version is 20x slower than this PR):

```
lint:
	@echo "\033[35mhttps://github.com/kisielk/errcheck\033[0m"
	errcheck ./app ./cfg ./flags ./help ./logger ./modules/... ./utils ./view ./wtf
	errcheck ./main.go

	@echo "\033[35mhttps://golang.org/cmd/vet/k\033[0m"
	go vet ./app ./cfg ./flags ./help ./logger ./modules/... ./utils ./view ./wtf
	go vet ./main.go

	@echo "\033[35m# https://staticcheck.io/docs/k\033[0m"
	staticcheck ./app ./cfg ./flags ./help ./logger ./modules/... ./utils ./view ./wtf 
	staticcheck ./main.go

	@echo "\033[35m# https://github.com/mdempsky/unconvert\033[0m"
	unconvert ./...
```

sam-github:
Yes, running the tool against multiple inputs simultaneously helps the current stuff a lot, its basically the optimization that golangci-lint does.

The current `lint` target assumes errcheck and staticcheck were installed globally, the `cilint` target assumes golantlint-ci is installed globally, its the same situation, isn't it? I'm pretty sure when I first tried to run `make lint`, I just looked up the name of the tools from the comments in the Makefile, and did a manual `got get ...` on them, then re-ran.

The build tool situation with go modules seems a bit obscure ATM, from https://github.com/golang/go/issues/25922 I've found and tried https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md, but I still needed a setup style target:

```
setup:
	GOBIN=$$PWD/bin go install github.com/golangci/golangci-lint

# Make the obvious lint target depend on setup, so it works "out of the box"
lint: setup just-lint

# If you have done setup already, don't repeat it, its slow
just-lint:
        ./bin/golangci-lint run 
```

Other than an explicit setup target, I don't see a way of ensuring dev-time tools are available. Adding a tools.go to the above seems to be a way to ensure exact versions of the tools are installed, which could be a big deal for a long-running large project that needs stable builds over years, but not so much here.

Makefile
```diff
@@ -86,6 +86,9 @@ install:
 	@echo "${APP} installed into ${INSTALLPATH}"
 
 ## lint: runs a number of code quality checks against the source code
+cilint:
```
sam-github, line 89: I should have mentioned, I called this `cilint` just so I could leave the original `lint` target in. Its a poor name, given that ci doesn't use it, it just uses the action. Oops.

I think the `cilint` target should be renamed `lint`, and the (current) `lint` target deleted.

senorprogrammer:
That’s an excellent point about the current one, I’d forgotten they’re not go standard. Let’s go straight with golintci.

sam-github:
Btw, I don't use vscode, but it has golangci-lint integrations: https://golangci-lint.run/usage/integrations/, might be useful (I'm not sure if the standalone tools also do that).

sam-github:
ok, give me a chance to tidy this up a bit, I'll add at least a couple comments so people know how to do the setup.

Actually, WDYT, `setup` target, or just a comment saying "go get it"? I suspect for vscode integration it might need to be global anyhow.

senorprogrammer:
Go with the simplest to implement option at this stage. I suspect 99% of contributors don’t use the `make` tooling so having it on CI is really what matters

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,20 @@
...
+      - master
+  pull_request:
+jobs:
+  golangci:
+    name: lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v0.2.0
```
jirfag, line 16: JFYI we've launched [v1](https://github.com/golangci/golangci-lint-action) with support of caching and without slow docker image pulling, it works much faster

Issue #907, opened 2020-05-27
PiHole 5.0 API update breaks PiHole module summary


jonhadfield:
As per title, a change to the API after upgrading from PiHole 4.x to 5.0 results in an error message being shown in place of the summary.
I'll try and submit a PR to make compatible with both versions this week.

senorprogrammer:
Fixed in https://github.com/wtfutil/wtf/pull/908

PR #908, opened 2020-05-28
make PiHole module compatible with both versions 4 and 5 of PiHole.

jonhadfield:
Fix for https://github.com/wtfutil/wtf/issues/907.
The issue was that the PiHole API in version 4 returns the relative blocklist age components as strings: 
`"relative":{"days":"0","hours":"00","minutes":"08"}`
whilst version 5 returns them as integers:
`"relative":{"days":0,"hours":7,"minutes":0}`
PR makes module compatible with both.

### DIFF
--- a/modules/pihole/client.go
+++ b/modules/pihole/client.go
@@ -23,9 +23,9 @@ type Status struct {
 	Status              string `json:"status"`
 	GravityLastUpdated  struct {
 		Relative struct {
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Awesome, thank you! Great to see this fixed (it will be nice to see the error message disappear from my dashboard)

PR #924, opened 2020-06-22
enable user to disable emoji for weather module.

jonhadfield:
On terminals that don't support emoji, the output of the weather status next to the title appears as a random character.
This PR enables a user to disable the output of that emoji.
The default behaviour is to have emoji as **enabled**. This is so that existing users are not affected and the assumption is that the majority of users' terminals will support emoji.


senorprogrammer:
Thanks!

PR #940, opened 2020-07-24
Close #876 - fix todo modal editing feeling slow

### LINKED ISSUE #876: Modal display when editing a Todo item is very slow
## What's broken?

Opening the modal for a new item is relatively fast. Opening the modal to edit an existing item is very slow. Figure out what the difference between them is. They should be the same speed (and that should feel fast).

aeter:
References #876. The modal shown when editing todos was feeling slow to show - there was required an additional hit of `Enter` to show the modal.

This commit  queues a `Draw` update in the `updateSelected` function (similarly to the `newItem` function) - so now the modal feels fast to show.

Tests:
* ran `make test`
* tested locally with a todo list


senorprogrammer:
Thank you for this. This will be out in the next release.

PR #944, opened 2020-07-29
fix: fix Jira module title formatting

### LINKED ISSUE #942: Bug: Jira module title formatting
## What's broken?

Screen Shot 2020-07-26 at 9 09 23 PM

My config (relevant parts):

```
    jira:
      title: " Jira"
      projects: ["Persona"]
```

Broken part in code: [modules/jira/widget.go#L80](https://github.com/wtfutil/wtf/blob/master/modules/jira/widget.go#L80)

Problem: `widget.settings.project` is an array, so gets stripped out during printing (I assume because it's treated as a color code).

(My personal suggested fix would just be to remove this extra stuff, it doesn't seem useful to me and the user can always put the project name in `title` if they want. Or maybe it can show issue count after the hyphen. Would be happy to file a PR either way!)

Nit 1: the code documentation says that the project(s) are passed as:

```
	projects                []string `help:"An array of projects to get data from"`
```

...but it is in fact [`project` and not `projects`](https://github.com/wtfutil/wtf/blob/master/modules/jira/settings.go#L66).

Nit 2: there is no whitespace between the word `Jira` and the hyphen, which I don't think looks good. (The Hackernews module has a space before and after the hyphen.)

timhwang21:
Closes #942. Filing to get the ball rolling, happy to change anything.

### DIFF
--- a/modules/jira/widget.go
+++ b/modules/jira/widget.go
@@ -77,7 +77,7 @@ func (widget *Widget) content() (string, string, bool) {
 		return widget.CommonSettings().Title, widget.err.Error(), true
 	}
 
-	title := fmt.Sprintf("%s- [green]%s[white]", widget.CommonSettings().Title, widget.settings.projects)
+	title := widget.CommonSettings().Title
 
 	str := fmt.Sprintf(" [%s]Assigned Issues[white]\n", widget.settings.common.Colors.Subheading)
 


timhwang21:
Derp, I'm bad at Go. Realized that the `settings` vs. `setting` thing was my mistake so reverted that.

senorprogrammer:
Thanks!

PR #976, opened 2020-10-01
Add two display modes for feedreader

cyingfan:
All 3 display modes:
- Title (current, default)
- Link
- Title + Content

The display mode is toggled using the  key `t`


senorprogrammer:
Thanks!

PR #981, opened 2020-10-03
Add natural ordering for clocks

### LINKED ISSUE #896: Clocks: sort in order specified
## What should it do?
```
clocks:
  locations:
    London: 'Europe/London'
    GMT: 'GMT'
```

Currently (with London on BST) there would be no way to get those displayed in the order specified. `sort: alphabetical` and `sort: chronological` will both put GMT first.

While asc/descending sort might be nice, what I'm actually requesting is `sort: as-written`, and hoping that that order is preserved by the yaml parser you're using (since it's not an array, I think the spec allows it not to be).

I'd go so far as to say 'as-written' sorting would be the most obvious default, too.

Seanstoppable:
This turns locations into an ordered array, so we can have a 'natural' ordering
This is backwards compatible
Resolves #896

### DIFF
--- a/modules/clocks/clock.go
+++ b/modules/clocks/clock.go
@@ -1,6 +1,7 @@
 package clocks
 
 import (
+	"strings"
 	"time"
 )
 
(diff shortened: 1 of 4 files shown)


senorprogrammer:
What would the config for this as a list look like?

Seanstoppable:
@senorprogrammer Here is the doc update. Essentially, anyone who wants to take advantage of this just prepends their lines with `-` to make it a YAML list.

https://github.com/wtfutil/wtfdocs/pull/100

senorprogrammer:
Awesome, thanks!

PR #982, opened 2020-10-03
Create generalized todo module

Seanstoppable:
Makes existing modules "backends"
This gives us a consistent interface


senorprogrammer:
Looks good, if the static checks can be fixed. This can be fixed by running `go mod tidy`.

Seanstoppable:
Thanks, just realized I was falling back to the original todoist library vs using the fork. Fixed! 🤞

senorprogrammer:
This is awesome.

PR #984, opened 2020-10-04
Fix config file path

cyingfan:
Background:
I noticed that my default config file (~/.config/wtf/config.yml) has the permission 666 but the expected permission is 600.
I traced down the issue to `cfg.chmodConfigFile()` where variable `relPath` carries the value `~/.config/wtfconfig.yml`.

Solution:
Use `filepath.Join()`


senorprogrammer:
Nice catch 👍

PR #985, opened 2020-10-05
Add info table tests

trimble:
In an attempt to assist with #657 , I offer a couple of very basic tests for info tables. Please recommend additional tests or other improvements and I'll update.


senorprogrammer:
Thanks! And anything that isn't already tested is fair game, and anything that is tested or can be improved or expanded is too. The more the better.

PR #988, opened 2020-10-07
Have cmdrunner use a tty

### LINKED ISSUE #577: cmdrunner should maintain color of output
## What's broken?

`cmdrunner` seems to be blowing away the color schemes of the outputs of many programs. 

## Reproduce

1. Choose a terminal program that outputs colored text. Example: `yarn info` or `jira issues ls` (from https://github.com/keepcosmos/terjira)
2. Create a new cmdrunner module and call the script.
3. All colors are lost.

System Version: macOS 10.14.5 (18F132)
Kernel Version: Darwin 18.6.0
GNU bash, version 5.0.2(1)-release (x86_64-apple-darwin18.2.0)

Seanstoppable:
This lets us get proper coloring
Fixes #577 

Looking at the below, it looks like something is not clearing the color correctly (likely another underlying library bug)

Raw command line: 
Screen Shot 2020-10-06 at 11 22 16 PM

CmdRunner  pre-commit:
Screen Shot 2020-10-06 at 11 20 27 PM

CmdRunner pre-commit, forcing color:
Screen Shot 2020-10-06 at 11 37 08 PM

CmdRunner post-commit:
Screen Shot 2020-10-06 at 11 21 44 PM

### DIFF
--- a/modules/cmdrunner/widget.go
+++ b/modules/cmdrunner/widget.go
@@ -125,22 +125,26 @@ func runCommandLoop(widget *Widget) {
 		cmd := exec.Command(widget.settings.cmd, widget.settings.args...)
 		cmd.Env = widget.environment()
 		f, err := pty.Start(cmd)
-		if err != nil {
-			panic(err)
-		}
-
-		io.Copy(widget.buffer, f)
 		// The command has exited, print any error messages
 		if err != nil {
-			widget.m.Lock()
-			_, writeErr := widget.buffer.WriteString(err.Error())
-			if writeErr != nil {
-				return
-			}
-			widget.m.Unlock()
+			widget.handleError(err)
+		}
+
+		_, err = io.Copy(widget.buffer, f)
+		if err != nil {
+			widget.handleError(err)
 		}
 		widget.redrawChan <- true
 	}
+}
+
+func (widget *Widget) handleError(err error) {
+	widget.m.Lock()
+	_, writeErr := widget.buffer.WriteString(err.Error())
+	if writeErr != nil {
+		return
+	}
+	widget.m.Unlock()
 }
 
 func redrawLoop(widget *Widget) {
--- a/modules/cmdrunner/widget.go
+++ b/modules/cmdrunner/widget.go
@@ -3,11 +3,13 @@ package cmdrunner
 import (
 	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"strings"
 	"sync"
 
+	"github.com/creack/pty"
 	"github.com/rivo/tview"
 	"github.com/wtfutil/wtf/view"
 )
@@ -121,10 +123,13 @@ func runCommandLoop(widget *Widget) {
 		<-widget.runChan
 		widget.resetBuffer()
 		cmd := exec.Command(widget.settings.cmd, widget.settings.args...)
-		cmd.Stdout = widget
 		cmd.Env = widget.environment()
-		err := cmd.Run()
+		f, err := pty.Start(cmd)
+		if err != nil {
+			panic(err)
+		}
 
+		io.Copy(widget.buffer, f)
 		// The command has exited, print any error messages
 		if err != nil {
 			widget.m.Lock()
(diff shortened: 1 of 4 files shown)


modules/cmdrunner/widget.go
```diff
@@ -121,10 +123,13 @@ func runCommandLoop(widget *Widget) {
 		<-widget.runChan
 		widget.resetBuffer()
 		cmd := exec.Command(widget.settings.cmd, widget.settings.args...)
-		cmd.Stdout = widget
 		cmd.Env = widget.environment()
-		err := cmd.Run()
+		f, err := pty.Start(cmd)
+		if err != nil {
+			panic(err)
```
senorprogrammer, line 129: I don't think we should panic here. If the `pty` fails for some reason, we should fall back to the old way to still provide command output.

senorprogrammer (changes requested):

modules/cmdrunner/widget.go
Seanstoppable: Thanks. Major brainfart on my behalf.

senorprogrammer:
This looks great. I'll try to get it in today.

sam-github:
Not sure if it matters, but pty is unlikely to work on windows, which doesn't have them. I'm not sure (I don't have windows :-),  maybe the pty module builds and gently errors out, instead of just failing to compile?

Doesn't have to be now, but running in a (pseudo)tty should perhaps become optional. Some commands change their behaviour based on whether they are in a tty or not, for example, they might assume that  a progress bar should be written if running in a tty because its a user running it interactively, but that might not be wanted if the command is running in wtf and only the output is wanted. Just a thought.

Seanstoppable:
@sam-github I hadn't thought about that. I added this mostly because if you look at the linked issue people were complaining that commands were not getting colorized be default, and they felt that was unexpected. Perhaps another option is to not even do this and just have additional wording around colorization in the docs.

senorprogrammer:
That's a great point, and it does matter. There are people using this on Windows and we should not break things on them needlessly. Make it opt-in via config?

sam-github:
I like colours, I think this solves a real problem, I didn't mean to suggest it wasn't useful.

Executables really do change their behaviour based on `isatty()`, so its great to be able to run them (or not, as needed) in a tty. Whether defaulting to being in a tty or not is right, I don't know. The plus side of in a tty is that the output would look like people see when trying the command in a shell, so it helps debugging, though for that, actually running in a shell would help, too (that's not related to this PR, though).

The downside is it might not be backwards compat, but I don't think wtf cares too much about that. I don't even think you should make it configurable if you don't have the interest, that can happen if someone actually has a problem, and wants to PR that themselves.

senorprogrammer:
"The downside is it might not be backwards compat, but I don't think wtf cares too much about that"

We care very much about backwards-compatibility :)

Seanstoppable:
The things we do for backwards compatability :)
I'll add a config option for it, as well as update the docs to try to make some of this clear
As the original ticket demonstrates, some things do have the option to force color, but presumably all things do not and may just make implicit choices.

modules/cmdrunner/widget.go
```diff
@@ -121,23 +123,45 @@ func runCommandLoop(widget *Widget) {
...
+
+	_, err = io.Copy(widget.buffer, f)
+	return err
+}
+
+func (widget *Widget) handleError(err error) {
+	widget.m.Lock()
+	_, writeErr := widget.buffer.WriteString(err.Error())
+	if writeErr != nil {
+		return
```
sam-github, line 160: Won't this leave the widget locked if WriteString fails?

senorprogrammer: Yeah, like L76 this should probably be `defer widget.m.Unlock()`

Seanstoppable: You are correct, issue with the original code. Will fix.

senorprogrammer:
Something strange is going on with this one. It's blocking my local test run and the Travis test run when it gets to the tests in `github.com/wtfutil/wtf/view`.

It doesn't look like it needs the `tview` upgrade, I think we should leave out of here (assuming that's the problem), and deal with that issue separately. That way this can get merged.

Seanstoppable:
Sounds good. I upgraded to see if it would help with coloring a bit, but didn't seem to have an impact.

senorprogrammer (approved):

senorprogrammer:
Nice!

PR #990, opened 2020-10-07
Add tests for specified min column width on info table

trimble:
In support of #657 , this adds tests for specifying minimum column widths on info table module columns.


senorprogrammer:
Nice!

PR #992, opened 2020-10-07
Upgrade golangci-lint action

Seanstoppable:
This hopefully gets rid of the deprecation warnings


Seanstoppable:
Actually, this won't solve the deprecation warnings, as there is an open issue there and the action itself doesn't have this code.

PR #995, opened 2020-10-08
Improve test coverage

jmks:
Re #657  

I haven't done much testing in Go, so that sounded interesting to try.

I tried to get [an example](https://golang.org/pkg/testing/#hdr-Examples) working but the colourized output seemed like a no go.

Anyways, I just split the `Validation` function into computing the error messages and printing them. The file's coverage ended up at 66%.

Also, when there is a config error, it printed two trailing new lines, so I changed it to just print one:

image

### DIFF
--- a/app/module_validator.go
+++ b/app/module_validator.go
@@ -5,10 +5,16 @@ import (
 	"os"
 
 	"github.com/logrusorgru/aurora"
+	"github.com/wtfutil/wtf/cfg"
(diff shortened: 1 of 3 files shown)


senorprogrammer:
The Travis error can be fixed by running `go mod tidy` and re-committing the result to the PR.

senorprogrammer:
Thanks!

PR #1000, opened 2020-10-13
Default domain for gitlab

jmks:
The [Gitlab module docs](https://wtfutil.com/modules/gitlab/) has the domain marked as optional.
But when I start it without one, I get `Get "/api/v4/user": unsupported protocol scheme ""`.

This makes the default domain `https://gitlab.com` which allows you to track open source projects like `gitlab-org/gitlab` or `inkscape/inkscape`.

Alternatively, the docs could be changed to mark the field as required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### 🐞 Fixed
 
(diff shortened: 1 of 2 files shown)


senorprogrammer:
Interesting. I wonder why this was working before without that being required.

senorprogrammer:
Thanks!

PR #1015, opened 2020-10-29
Handle errors from the grafana api by displaying the error message

schoentoon:
Stupidly enough only realized just now that it would show a vague json unmarshal error if the grafana api errors for whatever reason.


senorprogrammer:
Nice catch!

PR #1025, opened 2020-11-17
[CircleCI] Add configurable number of builds to settings

dabcoder:
Add option to set the number of builds in the config

### DIFF
--- a/modules/circleci/settings.go
+++ b/modules/circleci/settings.go
@@ -15,8 +15,8 @@ const (
 type Settings struct {
 	common *cfg.Common
 
-	apiKey      string `help:"Your CircleCI API token."`
-	buildNumber int    `help:"The number of build, 10 by default"`
+	apiKey         string `help:"Your CircleCI API token."`
+	numberOfBuilds int    `help:"The number of build, 10 by default"`
 }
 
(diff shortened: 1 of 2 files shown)


modules/circleci/settings.go
```diff
@@ -15,15 +15,17 @@ const (
 type Settings struct {
 	common *cfg.Common
 
-	apiKey string `help:"Your CircleCI API token."`
+	apiKey      string `help:"Your CircleCI API token."`
+	buildNumber int    `help:"The number of build, 10 by default"`
```
senorprogrammer, line 19: I would suggest `numberOfBuilds` instead. `buildNumber` could be misconstrued as a specific build number (ie. 453) rather than a quantity.

senorprogrammer (changes requested):

senorprogrammer:
Looks great to me. One small request and then I'm happy to merge it.

modules/circleci/settings.go
dabcoder: Agreed. Adressed this in 5b17ddf, thanks for the review. 
Not sure why the linting fails in https://github.com/wtfutil/wtf/pull/1025/checks?check_run_id=1421195730 though.

senorprogrammer:
That linting error is strange, and I'm going to ignore it.

senorprogrammer:
Good stuff!

PR #1034, opened 2020-11-30
Add krisinformation module

stone:
A module to parse the feed from: https://www.krisinformation.se/en

 Krisinformation.se is a web site run by the Swedish Civil Contingencies Agency
that compiles and convey warnings, alerts, and emergency information from
Swedish authorities to the public.

Features:

 - Filter on country (all messages)
 - Filter on county (only show county specific messages)
 - Distance (only show message within a certain distance)
 - Set a max age for displaying messages
 - Set a maximum number of messages displayed


senorprogrammer:
Nice work!

PR #1036, opened 2020-12-18
[CircleCI] Close the response body

dabcoder:
To avoid leaks, similar to what was done in another module: https://github.com/wtfutil/wtf/blob/cde904ff08897c0702de54d1dce0e5d21572bc0c/modules/bamboohr/request.go#L21.


senorprogrammer:
Great!

PR #1037, opened 2020-12-26
[modules] Add new covid tracker module

dabcoder:
This is a covid tracker module using this API: https://coronavirus-tracker-api.herokuapp.com/#/v2/. 

the config would look like:

```yaml
covid:
  enabled: true
  countries:
    - "FR"
    - "US"
    - "DE"
    - ...
   refreshInterval: 120
   position:
        ...
```

And the result would look like:

```
┌─────────────────────── Covid tracker ────────────────────────┐
│  ││Gobal
│  ││Confirmed: ...
│  ││Deaths: ...
│  ││Country: FR
│  ││Confirmed: ...
│  ││Deaths: ...
│  ││Country: US
│  ││Confirmed: ...
│  ││Deaths: ...
```

Remaining issue: the third country stats do not appear here, I tried setting `widget.View.SetScrollable` to true but that didn't seem to change anything.

### DIFF
--- a/modules/covid/widget.go
+++ b/modules/covid/widget.go
@@ -5,6 +5,9 @@ import (
 
 	"github.com/rivo/tview"
 	"github.com/wtfutil/wtf/view"
+
+	"golang.org/x/text/language"
+	"golang.org/x/text/message"
 )
 
 // Widget is the struct that defines this module widget
@@ -42,6 +45,14 @@ func (widget *Widget) Render() {
 	widget.Redraw(widget.content)
 }
 
+// Display stats based on the user's locale
+func (widget *Widget) displayStats(cases int) string {
+	prntr := message.NewPrinter(language.English)
+	str := fmt.Sprintf("%s", prntr.Sprintf("%d", cases))
+
+	return str
+}
+
 func (widget *Widget) content() (string, string, bool) {
 	title := defaultTitle
 	if widget.CommonSettings().Title != "" {
@@ -55,8 +66,8 @@ func (widget *Widget) content() (string, string, bool) {
 	} else {
 		// Display global stats
 		covidStats = fmt.Sprintf("[%s]Global[white]\n", widget.settings.Colors.Subheading)
-		covidStats += fmt.Sprintf("%s: %d\n", "Confirmed", cases.Latest.Confirmed)
-		covidStats += fmt.Sprintf("%s: %d\n", "Deaths", cases.Latest.Deaths)
+		covidStats += fmt.Sprintf("%s: %s\n", "Confirmed", widget.displayStats(cases.Latest.Confirmed))
+		covidStats += fmt.Sprintf("%s: %s\n", "Deaths", widget.displayStats(cases.Latest.Deaths))
 	}
 	// Retrieve country stats if country codes are set in the config
 	if len(widget.settings.countries) > 0 {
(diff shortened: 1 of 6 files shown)


senorprogrammer:
Could you describe the issue with the third country in more detail?  I'm running this locally and its looking good. To make the widget scrollable to see the rest of my data, I used this config:

```yaml
    covid:
      countries:
        - "FR"
        - "CA"
        - "US"
        - "DE"
      enabled: true
      focusable: true
      position:
        top: 3
        left: 2
        height: 1
        width: 1
      refreshInterval: 120
```

modules/covid/widget.go
```diff
@@ -0,0 +1,76 @@
...
+		title = widget.CommonSettings().Title
+	}
+
+	cases, err := LatestCases()
+	var covidStats string
+	if err != nil {
+		widget.err = err
+	} else {
+		// Display global stats
+		covidStats = fmt.Sprintf("[%s]Gobal[white]\n", widget.settings.Colors.Subheading)
```
senorprogrammer, line 57: sp `Global`

modules/covid/widget.go
```diff
@@ -0,0 +1,76 @@
...
+		covidStats += fmt.Sprintf("%s: %d\n", "Confirmed", cases.Latest.Confirmed)
```
senorprogrammer, line 58: I suspect that Confirmed and Deaths in this case should not be the same value.

modules/covid/widget.go
```diff
@@ -0,0 +1,76 @@
...
+	// Retrieve country stats if country codes are set in the config
+	if len(widget.settings.countries) > 0 {
+		countryCases, err := widget.LatestCountryCases(widget.settings.countries)
+		if err != nil {
+			widget.err = err
+		} else {
+			for i, name := range countryCases {
+				covidStats += fmt.Sprintf("[%s]Country[white]: %s\n", widget.settings.Colors.Subheading, widget.settings.countries[i])
+				covidStats += fmt.Sprintf("%s: %d\n", "Confirmed", name.Latest.Confirmed)
+				covidStats += fmt.Sprintf("%s: %d\n", "Deaths", name.Latest.Deaths)
```
senorprogrammer, line 70: For formatting numbers for easy readability, take a look at https://github.com/wtfutil/wtf/blob/master/modules/github/display.go#L126

It'll format the number according to the user's locale (in my case, like `554,282`). 

Fix these minor issues and I'd be happy to merge this PR in.

senorprogrammer (changes requested):

modules/covid/widget.go
dabcoder: Addressed in 46de9e53.

modules/covid/widget.go
dabcoder: Also addressed in 46de9e53.

modules/covid/widget.go
dabcoder: Gave it a try in 5934ed21, but I also hardcoded it to `language.English` like in the GitHub module. Is there a way to make this configurable, from the config yaml file maybe?

senorprogrammer:
Good point, I've removed the hard-coded English localization and made it configurable via a top-level `language` configuration key in this PR: https://github.com/wtfutil/wtf/pull/1038

Take a look here for how to implement the solution now: https://github.com/wtfutil/wtf/pull/1038/files#diff-1d37a88633fac3cd7f161204812dbf98d0267835aef4d6d5bc740a1bfe4163f7R125

senorprogrammer:
Merged your branch and updated the localization here https://github.com/wtfutil/wtf/pull/1037. You should be able to set yours in your config file now.

dabcoder:
> Merged your branch and updated the localization here #1037. You should be able to set yours in your config file now.

Thanks, much appreciated, will have a look and aim to do that in a subsequent PR.

> Could you describe the issue with the third country in more detail? I'm running this locally and its looking good. To make the widget scrollable to see the rest of my data, I used this config...

I used this config and didn't manage to display the third country stats:

```yaml
covid:
      enabled: true
      focusable: true
      countries:
        - "FR"
        - "US"
        - "DE"
      position:
        top: 1
        left: 2
        height: 1
        width: 2
      refreshInterval: 120
```

reportaman:
@dabcoder Thanks! Is it possible to specify state/province/region/city/town in the location for covid tracker?

dabcoder:
@reportaman sorry I missed your question. Just the country codes for now, this is the only option at the moment. You're right, I need to issue a PR for the documentation part.

PR #1040, opened 2021-01-01
[several modules] Close the response body

dabcoder:
Follow up of #1036, also systematically close the response body in some other modules.

### DIFF
--- a/modules/buildkite/client.go
+++ b/modules/buildkite/client.go
@@ -57,6 +57,8 @@ func (widget *Widget) recentBuilds(pipeline PipelineSettings) ([]Build, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer func() { _ = resp.Body.Close() }()
+
 	if resp.StatusCode < 200 || resp.StatusCode > 299 {
 		return nil, fmt.Errorf(resp.Status)
 	}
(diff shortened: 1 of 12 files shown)


dabcoder:
Test failure:

```
--- FAIL: TestFinnhubClient (0.60s)
panic: runtime error: index out of range [0] with length 0 [recovered]
	panic: runtime error: index out of range [0] with length 0
```

senorprogrammer:
Are you able to resolve this failure?

dabcoder:
> Are you able to resolve this failure?

@senorprogrammer I am not, no. Even if we close the response body at the end of the `finnhubRequest` function, the test fails with the same error. Any ideas what could be happening here?

senorprogrammer:
That might be, I'll take a look and see if I can figure it out.

senorprogrammer:
Looks good

dabcoder:
Thanks for following up on that one. Did you find out what the issue was?

senorprogrammer:
It was as you noted above, the test required a Finnhub API key. I've deleted the test here https://github.com/wtfutil/wtf/commit/9a703dd78b6737edea0d21d0f9a8ce72c2f6a8b0 since it did't add very much and ideally tests should not contact external services.

PR #1045, opened 2021-01-14
Healthcecks module

stone:
Healthchecks.io is a service for monitoring cron jobs and similar
periodic processes.

Hosted: https://healthchecks.io/
API-Doc: https://healthchecks.io/docs/api/
GitHub: https://github.com/healthchecks/healthchecks

This module can be used both with hosted and self-hosted healthchecks.


senorprogrammer:
Merged 👍

PR #1061, opened 2021-03-09
Fix city key

dabcoder:
Forgot to PR this a few weeks ago.


dabcoder:
cc @senorprogrammer when you have a moment.

senorprogrammer:
Pretty sure Avignon and Paris are in the same timezone but I don't mind this change :)

PR #1063, opened 2021-03-19
Fix HackerNews Widget - read on closed response body

LKaemmerling:
After commit 49109c77f719cedc5c7f15bcf255f952d2c3b90d the HackerNews widget did not work as expected. It failed with a error message "http: read on closed response body". This commit fixes the underlying reader issue by reading the whole request body within the apiRequest function and returning the []bytes of the response instead of *http.Response. If we want to continue to return *http.Response this would lead to the problem that the caller (GetStory() and GetStories()) always need to close the Request Body. The ways before this MR is a bit problematic because the defer gets executed when the function is "done", so it closes the Body already before the utils.ParseJSON can read it.

Signed-off-by: Lukas Kämmerling <lukas.kaemmerling@hetzner-cloud.de>

Screenshot before:
Bildschirmfoto 2021-03-19 um 06 30 44

Screenshot after:
Bildschirmfoto 2021-03-19 um 06 40 54

### DIFF
--- a/modules/hackernews/client.go
+++ b/modules/hackernews/client.go
@@ -1,7 +1,9 @@
 package hackernews
 
 import (
+	"bytes"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Nice patch.

Issue #1044, opened 2021-01-12
CircleCI Module


ejdunn2001:
I am getting an error: 
http: read on closed response body

senorprogrammer:
Thanks for filing this. I'll try and take a look.

yanivmo:
Seems like the deferred close, introduced in #1036, closes the body on return from `circleRequest`, before the body is actually read.

dabcoder:
@yanivmo thanks for the heads up, will look into this since I introduced it in the aforementioned PR. Interested in getting feedback with regards to where/when to actually close the response body if you already have a fix in mind.

LKaemmerling:
The same issue was/is living in the HackerNews module. I opened up an MR with a fix for this:
https://github.com/wtfutil/wtf/pull/1063

dabcoder:
@LKaemmerling Thanks for fixing this in the HN module, looks like we could do the same for this module too? Opened #1065 as a follow up.

senorprogrammer:
Looks like this has been fixed in the CircleCI module.

PR #1065, opened 2021-03-22
[CircleCi] Fix defer execution

dabcoder:
See #1044.


LKaemmerling (approved):
Looks good from my point of view 👍🏻 I can not test it as I don’t have a circleCI Account.

Issue #1088, opened 2021-06-03
feedreader module - Line breaks in titles


ajh-sr:
I'm not sure if this is a bug or if this is expected? Or if medium.com's feed is malformed? 

## What's broken?
![image](https://user-images.githubusercontent.com/71472057/120714076-da753980-c516-11eb-82f1-43ee0d3ff35a.png)

The article: [Top Contributors Impacting
Your Real-Time Application
Performance](https://subspacepowered.medium.com/top-contributors-impacting-your-real-time-application-performance-87e96676b2c7)

senorprogrammer:
Frankly, I'm surprised there are line breaks in a feed item title. That seems weird. But WTF should also handle that gracefully so I'm going to label this a bug and see about getting a fix for it.

GibranHL0:
Hey 👋, I'm new to the project and I'd like to contribute.
Could you assign me this issue?

senorprogrammer:
Great! It's all yours.

GibranHL0:
Thanks, I'll comment proposal as soon as possible.

ajh-sr:
🙇  Thanks @GibranHL0

senorprogrammer:
If you've got an idea for a fix, feel free to submit a PR. No need to propose, it's not that formal here :)

GibranHL0:
Opened a PR #1090

senorprogrammer:
Fixed by https://github.com/wtfutil/wtf/pull/1090

PR #1090, opened 2021-07-04
1088 - Add strings.Replace to prevent line breaks

GibranHL0:
Hey 👋,

I discovered gofeed doesn't check for line breaks within each item title. Also, I searched for how to implement multiline titles in Medium but I didn't have luck because it is a little weird to use it, but some people apply it.

Then, my solution was to replace all line breaks '\n' before it is presented on the widget.

I ran all tests as well.

### DIFF
--- a/modules/feedreader/widget.go
+++ b/modules/feedreader/widget.go
@@ -51,13 +51,14 @@ func rotateShowType(showtype ShowType) ShowType {
 }
 
 func getShowText(feedItem *FeedItem, showType ShowType) string {
-	returnValue := feedItem.item.Title
+	title := strings.Replace(feedItem.item.Title, "\n", " ", -1)
+	returnValue := title
 	switch showType {
 	case SHOW_LINK:
 		returnValue = feedItem.item.Link
 	case SHOW_CONTENT:
 		text, _ := html2text.FromString(feedItem.item.Content, html2text.Options{PrettyTables: true})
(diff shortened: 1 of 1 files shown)


modules/feedreader/widget.go
```diff
@@ -51,13 +51,14 @@ func rotateShowType(showtype ShowType) ShowType {
 }
 
 func getShowText(feedItem *FeedItem, showType ShowType) string {
-	returnValue := feedItem.item.Title
+	title := strings.Replace(feedItem.item.Title, "\n", " ", -1)
```
ajh-sr, line 54: Thanks for your PR to fix my issue! Apologies, haven't had a chance to test this yet, but could this lead to multiple spaces if the original title has a space either before or after the line break?
E.g. 
( where _ is a space character)
```
This is a title with_
a break
```

Might become:

```
This is a title with  a break
```

GibranHL0: Hey Adam 👋,

Yes, you are totally right, if the title has spaces and `\n` characters it will create an even spacier string.

Then, I came up with another solution using regex expressions. 

I committed those changes! 👀

ajh-sr: Awesome! 👍 

image
💚

ajh-sr (approved):
Looks good, and tested the PR on the original feed with the issue and it works for me. 👍 

ka pai!

senorprogrammer:
Closes https://github.com/wtfutil/wtf/issues/1088

PR #1095, opened 2021-08-02
Use HealthChecks API URL from settings when retrieving secret

hjr265:
This allows one to use the secret store with a self-hosted HealthChecks instance.

Without this, you are required to always save the secret against a specific URL: "https://hc-ping.com/".


sam-github (commented):
Looks reasonable to me.

senorprogrammer:
Looks good to me 👍

PR #1100, opened 2021-08-24
modules/todo enhancements

tadeaspaule:
Hi,

I find the todo module to be really useful, but it was missing some functionality so added it here:

- can move item to front of the list with Ctrl+F
- can move item to back of the list with Ctrl+L
- can set in yaml config whether new items are put as first or last with `newPos` field (valid values "first" or "last")
- can set in yaml config whether checked items are shown first or last or not specially resorted with `checkedPos` field (valid values "first", "last", "none")

Also date functionality:

- can specify date of todo with various prefixes, for example "in 3 days do this", "08-31 do something on the 31st of Aug", "tomorrow do X", "next week do Y"
- list is sorted according to dates, with undated todos counting as 7 days from today by default (changable by `dates.undatedAsDays` setting)
- can turn this whole functionality on/off with `dates.enabled` (true by default)
- can choose various formats with `dates.format` setting (yyyy-mm-dd, yy-mm-dd, dd-mm-yy, dd-mm-yyyy, dd M yy, dd M yyyy)
- can hide year if it's current year with `dates.hideYearIfCurrent` (true by default)
- can set how when dates flip to showing "in X days" instead of the date stamp with `dates.switchToInDaysIn` (default 7)

2021-08-31-092930_687x370_scrot

### DIFF
--- a/modules/todo/settings.go
+++ b/modules/todo/settings.go
@@ -14,9 +14,10 @@ const (
 type Settings struct {
 	*cfg.Common
 
-	filePath  string
-	checked   string
-	unchecked string
+	filePath   string
+	checked    string
+	unchecked  string
+	newToFront bool
 }
 
 // NewSettingsFromYAML creates a new settings instance from a YAML config block
@@ -29,6 +30,7 @@ func NewSettingsFromYAML(name string, ymlConfig *config.Config, globalConfig *co
 		filePath:  ymlConfig.UString("filename"),
 		checked:   ymlConfig.UString("checkedIcon", common.Checkbox.Checked),
 		unchecked: ymlConfig.UString("uncheckedIcon", common.Checkbox.Unchecked),
+		newToFront: ymlConfig.UBool("newToFront", true),
 	}
 
 	return &settings
--- a/modules/todo/settings.go
+++ b/modules/todo/settings.go
@@ -40,10 +40,10 @@ func NewSettingsFromYAML(name string, ymlConfig *config.Config, globalConfig *co
 	default:
 		settings.newPos = "last"
 	}
-	switch settings.checked {
+	switch settings.checkedPos {
 	case "first", "last", "none":
 	default:
-		settings.checked = "last"
+		settings.checkedPos = "last"
 	}
 
 	return &settings
--- a/modules/todo/settings.go
+++ b/modules/todo/settings.go
@@ -14,17 +14,17 @@ const (
 type Settings struct {
 	*cfg.Common
 
-	filePath   string
-	checked    string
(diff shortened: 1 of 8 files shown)


senorprogrammer:
Awesome! I’m on vacation right now but as soon as I’m back I’ll merge this in.

tadeaspaule:
Wonderful, hope you have a nice vacation :)

tadeaspaule:
Updated with one more configuration: can set the placement of checked items, so instead of shown at the end (this is still the default), you can use `checkedPos` in yml to also specify `first` if you want them to be shown first, or `none` (or anything else really) to not let checked status affect list sorting.

I find this really useful when you want to use the todo module as a quasi-Kanban board, I use checked items to indicate "in-progress", and want them to be shown (& moved to) the top of the list

modules/todo/settings.go
```diff
@@ -29,6 +37,30 @@ func NewSettingsFromYAML(name string, ymlConfig *config.Config, globalConfig *co
...
+	}
+	switch settings.checkedPos {
+	case "first", "last", "none":
+	default:
+		settings.checkedPos = "last"
+	}
+	switch settings.dateFormat {
+	case "yyyy-mm-dd", "yy-mm-dd", "dd-mm-yyyy", "dd-mm-yy", "dd M yy", "dd M yyyy":
+	default:
+		settings.dateFormat = "yyyy-mm-dd"
```
senorprogrammer, line 63: The only sensible date format :)

senorprogrammer (approved):

senorprogrammer:
@tadeaspaule Some linting issues to resolve and there are a few commented-out debugging lines to remove. Once those are resolved, I'll merge this in.

This is a great improvement - thanks for improving this!

tadeaspaule:
OK, everything should be in order now hopefully :) ran `gofmt` so lints should pass, noticed & fixed some minor issues, namely:

1.now resorts whole list at startup based on dates (only if `dates.enabled`). This takes care of `dates.undatedAsDays` situation changing (either config, or todo that was previously after undated todos should now be before it). @senorprogrammer Curious to hear your opinion on whether the whole list resort should happen more frequently? at the moment only the edited/new item is resorted, and it is assumed that the rest of the list is sorted (again, only if `dates.enabled`). This way there's less operations

2. move to first/last now synergises with dates (if `dates.enabled`): i.e. only moves to first/last of the same date (so if you have 10 todos for tomorrow / 10 todos that are undated, it moves to first/last of those 10). If `dates.enabled == false`, this happens as expected and moves to the very front/back of list

PR #1102, opened 2021-09-06
minor modules/git improvements through more config options

tadeaspaule:
Added the following config fields for `modules/git`:

```
sections         []interface{} `help:"Sections to show" optional:"true"`
showModuleName   bool          `help:"Whether to show 'Git - ' before information in title" optional:"true" default:"true"`
branchInTitle    bool          `help:"Whether to show branch name in title instead of the widget body itself" optional:"true" default:"false"`
showFilesIfEmpty bool          `help:"Whether to show Changed Files section if no changed files" optional:"true" default:"true"`
lastFolderTitle  bool          `help:"Whether to show only last part of directory path instead of full path" optional:"true" default:"false"`
```

2021-09-06-093322_312x247_scrot

### DIFF
--- a/modules/git/display.go
+++ b/modules/git/display.go
@@ -16,20 +16,44 @@ func (widget *Widget) content() (string, string, bool) {
 		return widget.CommonSettings().Title, " Git repo data is unavailable ", false
 	}
 
-	title := fmt.Sprintf(
-		"%s - %s[white]",
-		widget.CommonSettings().Title,
-		repoData.Repository,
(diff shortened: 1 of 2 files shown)


tadeaspaule:
Hi @senorprogrammer is this ok to merge?

(forgot to mention, I set the default values so that if the user doesn't change anything, everything looks as it did before this PR)

senorprogrammer:
Yep, looks great. Merged.

PR #1107, opened 2021-09-12
Tags and filter system

tadeaspaule:
2021-09-12-180517_701x347_scrot

- add tags to todos simply by typing `#tag` anywhere inside the text prompt when adding/editing an item
- filter items by tag by pressing `#` and entering a full tag or tag prefix to filter items by (for example you can enter `2` and only be shown todos tagged `#2read`, `2check`, etc). no need to enter the leading `#`
- filter items by substring by pressing `f` and entering text. items are shown only if they have this substring somewhere inside of them (case-insensitive)
- when you are filtering and press escape, the filter is wiped so you can see everything once again

you can modify the tag system a bit inside the yaml config:
- `tags.enabled` bool to enable/disable the whole system (default `true`)
- `tags.pos` is `start` or `end`, determines whether to display tags before or after todo text (default `end`)
- `tags.hide` list of tags to hide by default (so you have to use `#` to see them)
- `tags.hiddenInTitle` bool for whether to display number of items currently hidden by the tag + filter systems (default `true`)
- `colors.tags` to change the color of the tags when displayed (default `khaki`)

some examples of valid new/edit inputs that add tags correctly:

- `tomorrow laundry #chore` (if you want to add a date to the todo, it must come first!) -> text: `laundry`, tags: `[chore]`
- `#chore laundry` -> text: `laundry`, tags: `[chore]`
- `#this also #works` -> text: `also`, tags: `[this, works]`

### DIFF
--- a/view/keyboard_widget.go
+++ b/view/keyboard_widget.go
@@ -9,7 +9,7 @@ import (
 	"github.com/wtfutil/wtf/utils"
 )
 
-const helpKeyChar = "?"
+const helpKeyChar = "/"
 const refreshKeyChar = "r"
 
 type helpItem struct {
--- a/view/keyboard_widget.go
+++ b/view/keyboard_widget.go
@@ -9,7 +9,7 @@ import (
 	"github.com/wtfutil/wtf/utils"
 )
 
-const helpKeyChar = "/"
+const helpKeyChar = "?"
 const refreshKeyChar = "r"
 
 type helpItem struct {
--- a/checklist/checklist.go
+++ b/checklist/checklist.go
@@ -28,10 +28,11 @@ func NewChecklist(checkedIcon, uncheckedIcon string) Checklist {
 
 // Add creates a new checklist item and adds it to the list
 // The new one is at the start or end of the list, based on newPos
-func (list *Checklist) Add(checked bool, date *time.Time, text string, newPos ...string) {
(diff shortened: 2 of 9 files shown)


view/keyboard_widget.go
```diff
@@ -9,7 +9,7 @@ import (
 	"github.com/wtfutil/wtf/utils"
 )
 
-const helpKeyChar = "/"
+const helpKeyChar = "?"
```
senorprogrammer, line 12: I like everything about this, except this change. We can't change fundamentals like this on people without a really good reason. It'll break people's expectations and muscle memory.

Without this, I'm happy to merge this PR in.

senorprogrammer (changes requested):

view/keyboard_widget.go
tadeaspaule: You're right, sorry about that. I've been sick the past couple days, but I'll fix this up once I'm better + one small issue I noticed also needs to be done. Should be hopefully tomorrow or day after

senorprogrammer: Sounds good. Feel better!

tadeaspaule:
Hi, I fixed the things I mentioned but the checks are failing due to some `msgNoBattery` variable, which I see is in `master`. You can rerun checks when it's a good time for you and this time there should be no issues :)

senorprogrammer:
This is great, nice work.

PR #1109, opened 2021-09-26
fix nav and row color errors caused by tag/filter PR

tadeaspaule:
Hi, sorry but after the tag filter PR was merged I noticed issues with navigation and row highlighting caused by that PR. This fixes that

tldr: hiding rows didn't play too well with the usual system of getting row color based on index (since on the one hand the index is used for the nice alternating colors (just `% 2` check), so there you want to substract `hidden`, but elsewhere it's used and compared to `widget.Selected`)


senorprogrammer:
Awesome, thanks for catching and patching that so quickly.

PR #1111, opened 2021-10-01
Add Airbrake module

kyrylo:
This is a basic integration with [Airbrake](https://airbrake.io).

It only deals with errors and ignores other parts of Airbrake as APM, Deploys
and such.

The view shows 10 most recent errors. The list updates regularly.

You can view, resolve, unmute and mute errors. You can also open a certain error
in your browser. There's also an additional "compare view" available. It allows
you comparing error and comment counts.


senorprogrammer:
This will be great to add in. Just needs some cleanup on the linting.

kyrylo:
I fixed the golangci-lint errors, but I am not sure if there's anything else left to fix because the checks are not running without your approval. It also looks like the history of the old checks is gone now that I force-pushed.

PTAL

kyrylo:
I think I got rid of all the lint errors.

PTAL again.

Issue #1016, opened 2020-11-02
Can I have a Google Calendar with two email address?


fakeharxy:
## How can we help?

I apologise if this has been asked before. I have looked. There is quite a lot of talk about getting multiple calendars from one gmail account - but I have a gmail for work and a gmail for home. It would be great to combine the two calendars, or have two instances of `gcal:` running next to each other - but if I do that, the second one seems to overwrite the first. Is there any way of doing this.

Thanks!

senorprogrammer:
There's no support for multiple emails in the same calendar view (ie no way to merge the data from two distinct calendars). That sounds like it would be a good feature to add.

You can run multiple instances of the calendar widget by configuring the two of them uniquely. Take a look at the docs here to see how to do that: https://wtfutil.com/configuration/add_mutiple_instance_of_widget/

fakeharxy:
Ah! That's precisely what I wanted, thank you.

fakeharxy:
Ha, sadly, this is so close but no cigar. Because the gcal-auth.json file cannot be shared, and as far as I can tell, you can't set the file name - you can for the secret but not the response file. Maybe I'll learn Go and see about merging those two calendars after all ;)

senorprogrammer:
That is an excellent point. Feel free to take on the multiple calendars task if you like, and ask in here or the Slack channel if you need any pointers. I'm going to extract that file path out so that it's configurable.

pranavprabhakar:
@fakeharxy I'm looking for the exact same feature. Any update on this?

PR #1112, opened 2021-10-03
Allow multiple gcal widgets

Seanstoppable:
Store cached credentials referenced by email
This solves #1016


Seanstoppable:
Note: This WILL break backwards compatibility (force people to re-login).

Any thoughts on how to mitigate/migrate?

senorprogrammer:
When you say they'll need to log back in, what does that look like? I haven't changed my GCal settings since I first build the module - I forget what that process is like!

PR #1114, opened 2021-10-04
Add `personalAccessToken` setting to jira module

### LINKED ISSUE #1060: Jira auth with PAT
## What should it do?

Currently, I have to use username : password authentication with basic auth.

It would be nice to also accept a personal access token in order to not having to expose full login information.

PTAs can be used via the Authorization header, see https://confluence.atlassian.com/enterprise/using-personal-access-tokens-1026032365.html#UsingPersonalAccessTokens-UsingPATs

avdv:
When a PAT is set, it is used instead of the username / password
based authentication.

Fixes #1060

### DIFF
--- a/modules/jira/client.go
+++ b/modules/jira/client.go
@@ -63,7 +63,11 @@ func (widget *Widget) jiraRequest(path string) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


Seanstoppable:
Looks great, thanks!

PR #1116, opened 2021-10-05
Upgrade tview/tcell to latest

Seanstoppable:
This is likely a precursor to migrating to cview


senorprogrammer:
[cview](https://code.rocketnine.space/tslocum/cview) looks really interesting. I think a migration to that makes sense. Any thought on how much effort that would be?

Seanstoppable:
Not much effort at all!
Slightly more involved than this PR (renaming all the imports), and there are a few small code changes as well.

Issue #1103, opened 2021-09-07
refreshInterval less than 1 second


reportaman:
I know this might sound crazy for a dashboard, but some modules can happily take a refresh rate of less than 1 second. It can be a stopwatch for instance...

noxer:
Maybe this could be implemented by adding an 'RefreshUnit' field to the common config and then multiplying 'RefreshInterval' with it.

senorprogrammer:
Before tackling this problem at all, I'd like to see a few valid use-cases for a faster refresh rate that have a practical application. I'm not yet convinced this is a problem worth solving.

senorprogrammer:
Solved by @ResamVi. Configurations now support an optional time unit, ie: `50ms, 5s`, etc. A single integer will still be interpreted as seconds.

reportaman:
@senorprogrammer @ResamVi Thanks, works well!!!

PR #1120, opened 2021-10-09
Implement refreshIntervals < 1 seconds by adding refreshUnit field

ResamVi:
Solves #1103

Adds a `refreshUnit` field to common config (as proposed by @noxer)
Adds a (slow running) test as a sanity check.

### DIFF
--- a/cfg/parsers.go
+++ b/cfg/parsers.go
@@ -28,7 +28,7 @@ func ParseAsMapOrList(ymlConfig *config.Config, configKey string) []string {
 	return result
 }
 
-// ParseTimeString takes a configuration key and attempts to parse it first as a int
+// ParseTimeString takes a configuration key and attempts to parse it first as an int
 // and then as a duration (int + time unit)
 func ParseTimeString(cfg *config.Config, configKey string, defaultValue string) time.Duration {
 	i, err := cfg.Int(configKey)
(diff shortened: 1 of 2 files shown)


reportaman:
@ResamVi Thanks! There are times when a user might want to observe some data of interest from second to second. For some users this can be observing network utilization, and having the screen update faster than 1 update per second would be useful.

senorprogrammer:
I've reverted this change because something about it causes WTF to hang, preventing it from quitting or refreshing.

ResamVi:
So what's happening is that the `textfile` module is not using the `Schedule` function like every other module does. 
Instead it deploys it's own scheduling logic  in `watchForFileChanges` because - unlike every other module - it refreshes on file changes instead of regular intervals.

That logic however didn't listen to `ctrl+c` and wouldn't properly exit.
So I simply added the below case for it to work. Now the WTF properly exits.
```diff
func (widget *Widget) watchForFileChanges() {
	watch := watcher.New()
	watch.FilterOps(watcher.Write)

	go func() {
		for {
			select {
			case <-watch.Event:
				widget.Refresh()
			case err := <-watch.Error:
				fmt.Println(err)
				os.Exit(1)
			case <-watch.Closed:
				return
+			case quit := <-widget.QuitChan():
+				if quit {
+					return
+				}
			}
		}
	}()
```

Changing from `int` to `time.Duration` somehow unearthed this problem and I don't really know why.

However: Can someone please test what the end screen shows? I'm getting different results in VSCode's Terminal
vs. a normal terminal and I'm not sure if I am the cause:

hmmm

PR #1128, opened 2021-10-13
fix make lint and make gosec error (#1127)

### LINKED ISSUE #1127: make lint / make gosec failed if golangci-lint / gosec is not installed
## What's broken?

```sh
$ make lint
golangci-lint cache clean
make: golangci-lint: No such file or directory
make: *** [Makefile:105: lint] Error 127
```

yosmoc:
fix #1127

Install golangci-lint / gosec, if it is not installed.
If we only support over go 1.16, we should use go install instead of go
get.

Use binary installation for golangci-lint because they recommends.
https://golangci-lint.run/usage/install/#local-installation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,8 @@ export GOPROXY = https://proxy.golang.org,direct
 # Determines the path to this Makefile
 THIS_FILE := $(lastword $(MAKEFILE_LIST))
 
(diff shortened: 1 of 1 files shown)


senorprogrammer:
We do only officially support Go 1.16 and greater (with a preference for 1.17.x)

Issue #1049, opened 2021-01-23
Enable horizontal scrolling in git module


reportaman:
## What should it do?

Please enable horizontal (left-right) scrolling in git module just like other modules that have both vertical (up-down) as well as horizontal scrolling. Currently git module only has vertical scrolling enabled.

This will enable space utilization in smaller terminal windows

PR #1129, opened 2021-10-13
Support horizontal scroll in Git module (#1049)

yosmoc:
This PR is just my suggestion for #1049.

In order to support horizontal scrolling, we need to think of key bindings.

Originally, left/right key are assigned to move the projects.
h/l keys are also assigned to this function.

I suggest changing left/right key for horizontal scrolling and h/l keys for moving the project.
I thought to use shift as a modifier key, but other module doesn't use this kind of complicated key bindings.

### DIFF
--- a/modules/git/keyboard.go
+++ b/modules/git/keyboard.go
@@ -1,7 +1,5 @@
 package git
 
-import "github.com/gdamore/tcell"
-
 func (widget *Widget) initializeKeyboardControls() {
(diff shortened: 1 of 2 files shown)


senorprogrammer:
I like the idea, but not the key remapping. That introduces a breaking change that's non-standard (other modules also use arrow keys to move between items). I propose using `[` and `]` for explicit horizontal scrolling. What do you think?

PR #1131, opened 2021-10-15
Removed .travis.yml

Choromanski:


Choromanski:
Didn't think this was still being used.

senorprogrammer:
Neither did I. I thought I'd removed Travis from the pipelines. What did this removal break?

PR #1132, opened 2021-10-16
Fix digitalocean Image / Region information bug

### LINKED ISSUE #1053: DigitalOcean module: region does not show up
## What's broken?

When displaying DigitalOcean droplets, the "Region" column shows "???" instead of the droplet region.
Screen Shot 2021-02-03 at 1 10 03 PM

yosmoc:
Fix #1053 

Correct ref(address) is not passed to StringValueForProperty before.
Send correct ref as an argument to StringValueForProperty.

### DIFF
--- a/modules/digitalocean/droplet.go
+++ b/modules/digitalocean/droplet.go
@@ -1,8 +1,6 @@
 package digitalocean
 
 import (
-	"fmt"
-	"reflect"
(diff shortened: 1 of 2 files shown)


senorprogrammer:
This is a beautiful cleanup. Nice work.

Issue #957, opened 2020-08-23
Add basic authentication support for RSS feed reader


atgreen:
## What should it do?
From what I can tell, the feed parser can only read public feeds.  Please consider adding support basic authentication support for private feeds.

FliegendeWurst:
It is already possible to specify the username and password like this: `https://username:password@example.com/feed.rss`

senorprogrammer:
@atgreen Does this solution work for you?

ResamVi:
Hi, I'm looking around for hacktoberfest.
Adding username and password to the URL does do the trick, I've just tried it in a test.

Would it be useful if username+password attributes were added to the module?
That would be the solution a single private feed (multiple feeds with different credentials would require a bit more)

senorprogrammer:
I think extending the module config to support those values, and being able to use them if they're both supplied in the feed URL or in the configuration would be a great improvement.

senorprogrammer:
For that, might want to change the config format from a list to a map. We'd still need to support both for backwards compatibility (we do this in a number of places, try parsing one type first, if that fails, fall back to the default type and try again).

PR #1133, opened 2021-10-16
Add fields for basic auth

ResamVi:
Solves #957 

~~This adds a field `auth` to the feedreader module:
For private rss feeds returning a `401` it attempts basic auth
by cycling through the auth list~~ This was changed (see [comment below](https://github.com/wtfutil/wtf/pull/1133#issuecomment-954920898))

E.g.:
```yml
feedreader:
  enabled: true
  feeds:
  - http://localhost:3000/feed/rss.xml
  auth:
  - ["user", "password"]
  - ["john doe", "supersecret"]
  feedLimit: 10
  position:
    top: 1
    left: 1
    width: 2
    height: 1
  refreshInterval: 14400
```

I've added a test but it's in the `app` folder because otherwise it would cause an import cycle.

This upgrades the [gofeed dependency](https://github.com/mmcdole/gofeed) to a more recent commit (three days ago, that's some lucky timing) where [basic authentication is implemented](https://github.com/mmcdole/gofeed/pull/176)


Seanstoppable:
I'm concerned about the strategy of cycling through the auth list, as you are effectively sending your credentials to a bunch of sites.
Per @senorprogrammer's comment on the issue, we probably want a more robust struct of data if we want to support this, so there is a 1:1 mapping of credentials to url.

senorprogrammer:
Looks great 💯

PR #1139, opened 2021-10-29
Refreshintervals less than one second

ResamVi:
Solves #1103
In continuation of #1120

This feature made `wtf` hang up and couldn't be exit with `ctrl+c`:

 So what happened is that the `textfile` module is not using the `Schedule` function like every other module does. 
Instead it deploys it's own scheduling logic  in `watchForFileChanges` because - unlike every other module - it refreshes on file changes instead of regular intervals.

That logic however didn't listen to `ctrl+c` and wouldn't properly exit.
So I simply added the below case for it to work. Now WTF properly exits.
```diff
func (widget *Widget) watchForFileChanges() {
	watch := watcher.New()
	watch.FilterOps(watcher.Write)

	go func() {
		for {
			select {
			case <-watch.Event:
				widget.Refresh()
			case err := <-watch.Error:
				fmt.Println(err)
				os.Exit(1)
			case <-watch.Closed:
				return
+			case quit := <-widget.QuitChan():
+				if quit {
+					return
+				}
			}
		}
	}()
```

Changing from `int` to `time.Duration` somehow unearthed this problem and I don't really know why.

I've changed all instances of `refreshInterval int` to `refreshInterval time.Duration` in every module to keep it completely consistent and not have any loose ends open.


senorprogrammer:
Awesome work 🎉

Issue #1082, opened 2021-04-17
Module: Exchange Rates - Don't work any more


KubikPixel:
The Module Exchange Rates don't work anymore, shows only 0 (zero). What I see and I expect is, the API for the exchange rates are changed and needs now a API key and allow only in the free version 250 requests/mnt.

I don't know but it es evt. the best solution to change the API provider.

senorprogrammer:
Confirmed. Just seeing zeros.

senorprogrammer:
Looks like there have been fundamental changes to the API by the provider. Given that I don't use this module I will likely remove it and if someone else wants to create an exchange rates module using a better provider, they can.

Ferenc-:
A few APIs which might likely be around a bit longer:
* https://query1.finance.yahoo.com/v7/finance/quote?lang=en-US&region=US&corsDomain=finance.yahoo.com&fields=regularMarketPrice&symbols=EURUSD=X
* https://www.ecb.europa.eu/stats/eurofxref/eurofxref-daily.xml
* https://www.mycurrency.net/=US

Should we refurbish the existing exchange module, or create a new one?
(For example the yahoo API can do more than just currency.)

Ferenc-:
Actually the recently introduced https://github.com/wtfutil/wtf/tree/master/modules/stocks/yfinance can already do currency exchange. Perhaps the documentation could be more clear there.

senorprogrammer:
That's interesting. Do you have an example of how to configure it for that? I don't use this module but if you can provide one, I'll update the docs accordingly.

Ferenc-:
```
    yfinance:
      title: "💰 Rates"
      symbols:
        - "EURUSD=X"
        - "USDCAD=X"
        - "CADJPY=X"
      sort: true
      enabled: true
      refreshInterval: 60
      position:
        top: 0
        left: 5
        height: 1
        width: 1
```

PR #1150, opened 2021-11-04
Drop support exchangeratesapi.io

yosmoc:
Related #1082

Use stocks/yfinance module instead
exchangeratesapi.io has also currency convert api, but this api is just
for a paid plan. The free plan cannot use this api.


senorprogrammer:
This is a good idea.

Issue #1138, opened 2021-10-22
Module feature (Feed Reader): Add a label to display, and perhaps also group by it


reportaman:
## What should it do?

The feeds of the feedreader are probably randomized. It would be nice to see which feed is coming from which RSS feed source by seeing a label associated with that source. Example...

Setting:
```yaml
      feeds:
        - https://www.reddit.com/r/linux/.rss
        - https://www.reddit.com/r/neovim/.rss
        - https://www.reddit.com/r/vim/.rss
        - https://www.reddit.com/r/zsh/.rss
        - https://www.reddit.com/r/bash/.rss
```

returns something like this:
```
 1. [MATE+ Compiz] this graphical glitch is happening on m
 2. Quick tip for anyone having CPU spikes with EFM       
 3. Any idea what might be causing my Terminal to look lik
 4. Python Script Workaround to Be Able to Use Virtual Bac
 5. nvim-lsp-typescript-config                            
 6. Suggest some plugins and user experience              
 7. Lualine now has a new home                            
 8. lspsaga best fork?                                    
 9. clear fzf.vim unexcepted highlight column             
10. i need a vimscript wizard again                       
11. Clementine vs Strawberry media players                
12. How To Access A Particular Field In An Index In An Arr
13. Nvim lsp javascript formating on save issue with open 
14. PSA: vim.diagnostics api has changed a little bit on m
15. MX Linux 21 has been released                         
16. onedarkpro.nvim - cause the community needs more oneda
17. How do I get sed to remove a string that contains cont
18. Autosaving manual folds not working                   
19. How Much Control will Users have over Scheduling on Al
20. Is there a way to stay in command mode after pressing 
```

This ticket is asking for user to be able to specify a label for each feed source, and append it in front of the feed OR with an option ``grouped`` group the feeds belonging to each rss feed label and present them together. For grouped mode, ideally color the label with a different color to make it stand out from the rest of titles of feeds under it.

Benefit: It would make it simple for user to look at the ares of their interest at any given point in time and jump accordingly. At present looking at the random feed requires scanning all the lines to find the ones belonging to present topic of interest.

igbanam:
So I saw this and thought to take a stab at this. Please review the above and let me know if there is any better way I should be writing Go.

p.s: This is the first time I looked into Go ever. So I am winging it on half a wing here. Looking to learn from y'all on Go best practices.

PR #1171, opened 2022-01-13
Tag Stories in FeedReader

igbanam:
With the FeedReader, it's possible to have multiple feeds. With multiple feeds, all stories are jumbled up in the same list. One cannot know the source of a story at first glance. The way this possible is the to use the `t` keyboard shortcut which cycles the showType through title, link, and content. It is better UX to have some tag showing the origin of a story.

This change introduces a tag on the story title. This tag is the same as the Feed.Title. This change is backward compatible. If a feed does not come with a title, this change yields itself to the previous implementation.

To test this, run tool with `make run`. Feeds which have titles in their schemas should show up with their titles in square brackets. Examples of such feeds I have in my config are Hacker News, Ruby Weekly, and avdi.codes

Fixes wtfutil/wtf#1138

### DIFF
--- a/modules/feedreader/widget.go
+++ b/modules/feedreader/widget.go
@@ -24,8 +24,9 @@ const (
 
 // FeedItem represents an item returned from an RSS or Atom feed
 type FeedItem struct {
-	item   *gofeed.Item
-	viewed bool
+	item		*gofeed.Item
+	sourceTitle	string
+	viewed		bool
 }
 
 // Widget is the container for RSS and Atom data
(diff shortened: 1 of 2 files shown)


igbanam:
Before…

Screenshot 2022-01-13 at 18 13 19

After…

Screenshot 2022-01-13 at 18 03 27

senorprogrammer:
I think your linter issues can be resolved by running `gofmt` on the `/feedreader/*` files. 

`❯ golangci-lint run` gives this output:

`File is not `gofmt`-ed with `-s` (gofmt)`

senorprogrammer:
This looks good to me. If you like, I can fix that linting issue and merge it in, or you can fix it. Let me know either way.

PR #1212, opened 2022-04-02
Updated Install Paths in Makefile

gcg:
### Makefile install paths updated

Switched static install paths from `~/go/bin` to `$(GOBIN)` so it works everywhere.


senorprogrammer:
Apologies for the delay. Good idea!

PR #1219, opened 2022-04-15
fix(twitch): oauth token, stream title, title change, upgrade twitch lib

despreston:
A while ago twitch changed their GET /streams route and it now requires
an oauth token. This change adds a few more config options to support
this. If a client secret is given and an access token is not, an oauth
token will be fetched automatically. The token is refreshed each time
the module is refreshed to ensure an expired token is never used. This
was a simpler solution than tracking the expiration date given in the
oauth token.

This also changes the title of the module. "Top Streams" was kind of
ambiguous.

Upgrading the library used to access the twitch API to v2.


despreston:
Little stumped by what your static checker is asking for. I need to create new `Stream` structs somehow. What I have seems to be the simplest way to do that but maybe you could spare some advice?

senorprogrammer:
Yeah, the linters seem pretty fundamentally broken today. Not sure why.

senorprogrammer:
This looks good and will be merged in once the breaking changes in master are resolved (https://github.com/wtfutil/wtf/issues/1221).

As for the DeepSource issue, I'll be able to take a look next week, but at first glance I think you're right.

PR #1237, opened 2022-05-04
Update tickets.go

1waterrj:
tickets.go:68: json: cannot unmarshal number 398485867911 into Go struct field Ticket.Tickets.requester_id of type uint32

requester_id too big


Seanstoppable (commented):
Appreciate the PR!

Mind updating the following as well? I would expect them to hit similar issues:
SubmitterId
AssigneeId
CollaboratorIds

Imagine, essentially if all those were the same

Seanstoppable:
I don't think that is necessary, since it isn't code that you touched.

PR #1341, opened 2022-09-12
Twitch module - followed streams and launch via streamlink

zyrre:
I added necessary config to be able to view my followed twitch streams instead of just top streams. Also option to launch via streamlink (launches mpv to show the stream instead of the website: https://github.com/streamlink/streamlink).

It was necessary to add user access token and refresh token in order to achieve this, as twitch API has some different tokens. To get top streams you only needed app access token.

Also added config value to select which streams to list, "followed" or "top"(default).

I can probably add some steps to the twitch module documentation after this, as I just had to wrestle with the twitch API with how to get the correct tokens :)


senorprogrammer:
If you can fix that linting issue (catch and respond to the err value), I'll merge this in.

zyrre:
Ah, missed that one. I added checking for it and passing to handleError().

zyrre:
Also added PR for docs https://github.com/wtfutil/wtfdocs/pull/154

PR #1352, opened 2022-09-20
Add mempool recommended transaction fee module

Bizzaro:
# What 
- Add new mempool module.
- Rename `cryptoexchanges` -> `cryptocurrency` so everything is in one place.

# Docs 
https://github.com/wtfutil/wtfdocs/pull/156


Bizzaro:
@senorprogrammer this is ready now, please take a look when you get a chance.

PR #1355, opened 2022-09-20
Add nextbus API integration from umoiq

Bizzaro:
# What
- Adds next bus prediction from NextBus API
- http://www.nextbus.com/xmlFeedDocs/NextBusXMLFeed.pdf
- https://gist.github.com/Bizzaro/1889b61117bdff824210beebe1b2f868

# Docs
https://github.com/wtfutil/wtfdocs/pull/155

# Known issues
I learned `go` over the weekend so I'm not sure of a cleaner way to unmarshal the JSON object. In `javascript`, this takes no time because there is no strict typing, and seems like I'm used to first party support for parsing JSON 😭😭😭. With `go` this took me 2 days. The response returns an array when there's more than 1 entry and it returns an object when theres only 1... so if there's a better way, please feel free to suggest it!


Bizzaro:
@senorprogrammer this is ready now, please take a look when you get a chance.

PR #1378, opened 2022-10-10
Add instructions for go install to README.

winder:
Added instructions for installing the command directly with `go install`. I was taking a look at the project and this is how I installed it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,12 @@ and you should be good to go.
 
 ### Installing from Source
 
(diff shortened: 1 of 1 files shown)


senorprogrammer:
Good idea!

senorprogrammer:
If you look at the instructions right below the ones you wrote, they do that, and more. Your way certainly works for the common case, but the `make install` approach is more sanitary.

Issue #1342, opened 2022-09-14
Reddit module and feeds error with 403 Forbidden  


besttof:
## Reddit module and feeds error with 403 Forbidden  

Whenever I add a Reddit feed, either via the `feedreader` module or with the dedicated `subreddit` module, the result is always 403:
```
┌────── /r/news - top 3 ───────┐
│403 Forbidden                 │
│                              │
│                              │
│                              │
│                              │
│                              │
└──────────────────────────────┘
```

Config examples:
```
    subreddit:
      enabled: true
      numberOfPosts: 10
      refreshInterval: 15m
      sortOrder: top
      subreddit: "news"
      topTimePeriod: month
```
```
    feedreader:
      enabled: true
      feeds:
      - https://www.reddit.com/r/dwarffortress.rss
      feedLimit: 10
      refreshInterval: 4h
```

There are no issues using curl for that same rss feed or visiting reddit in the browser. However, if I try to do `curl https://www.reddit.com/r/dwarffortress.rss` after quitting wtfutil, it returns a `Too Many Requests` page. Which is weird, because the refresh intervals are very reasonable.

wtutil 0.41.0 (2021-12-08T06:06:22Z)
maxOS 12.3
iTerm2 3.4.16 (but system terminal yields the same results)

foreignsasquatch:
Having the same issue here on endeavour os

tsbkw:
I can reproduce the issue of feedreader but could not reproduce of subreddit with Config examples.

My environment is below. (v0.42.0)
```
% wtfutil -v
a63329214c888cfbfc67c7ddcf31887c3c8a1c36 (2022-10-05T15:08:35Z)
```

config (I edited position attributes of Confing examples)
```
    subreddit:
      enabled: true
      numberOfPosts: 10
      refreshInterval: 15m
      sortOrder: top
      subreddit: "news"
      topTimePeriod: month
      position:
        top: 4
        left: 1
        height: 1
        width: 2
    feedreader:
      enabled: true
      feeds:
      - https://www.reddit.com/r/dwarffortress.rss
      feedLimit: 10
      refreshInterval: 4h
      position:
        top: 5
        left: 1
        height: 1
        width: 2
```

output
```
                                ┌────────────────────── /r/news - top 3 ───────────────────────┐                                                                                                        
                                │ 1. Biden to pardon all prior federal offenses of simple marij│                                                                                                        
[...]

PR #1394, opened 2022-10-21
fix: Change subreddit user-agent to conform to reddit API rules

tsbkw:
When I investigate #1342 , I found current User-Agent used by subreddit is not conform to [reddit API rules](https://github.com/reddit-archive/reddit/wiki/API) . 
I try to change user-agent to conform to reddit API rules as much as possible.

This change add new dependency from flags to utils, in order to define function to get build version and time in consistant way. If it is not appropriate, I'll copy implementation in flags to subreddit (however this may cause inconsistent version in user-agent and -v option when version logic is changed ).


Seanstoppable:
So, I went and tested this, but still getting 'too many requests'.

Vs, what is currently merged to master, where that still seems to work.

Can you verify that the current master doesn't work, but this PR doesn't? I can try poking around again.

tsbkw:
@Seanstoppable
As I wrote I only changed to conform to reddit API rules to prevent problems in the future and this pull request doesn't stop error `too many requests`.

> I try to change user-agent to conform to reddit API rules as much as possible.

I also investigated on the problem and found feedreader occur `too many requests` error because of GoFeed user agent and this problem cannot be avoided because currently GoFeed doesn't provide the way to change user agent.

https://github.com/wtfutil/wtf/issues/1342#issuecomment-1286569429

Seanstoppable:
I had changed the user agent to conform closer to the rules to fix `too many requests`. 
I see what you are getting at here, but switching to this actually causes `too many requests` to occur again. Hence, this feels like a regression in functionality.

PR #1396, opened 2022-10-22
Add protocol version configuration to `ipinfo` module and update dependency

### LINKED ISSUE #1110: IPInfo module features: scrolling and option for ipv4 address
## What should it do?
Currently the IP info module shows ipv6 address in my dashboard, and I would like to see ipv4. So the request is to add an option so user can specify which address is shown. Interestingly, the address shows in the screenshot at https://wtfutil.com/modules/ipaddress/ipinfo/ is ipv4. So it seems to vary by user.

It would also be great to have a scrolling option, currently ipv6 address does not fit in the window width I had allocated to IPInfo module.

vadeg:
* Add `protocolVersion` to the `ipinfo` module to select IP protocol version to use
* Update `github.com/andybalholm/brotli` dependency. Otherwise, the project doesn't compile locally.

Fixes #1110

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -87,7 +87,7 @@ require (
 	github.com/PuerkitoBio/purell v1.1.1 // indirect
 	github.com/PuerkitoBio/urlesc v0.0.0-20170810143723-de5bf2ad4578 // indirect
 	github.com/acomagu/bufpipe v1.0.3 // indirect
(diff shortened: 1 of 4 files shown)


vadeg:
@senorprogrammer could you please have a look at this PR?

senorprogrammer:
Looks good to me

PR #1443, opened 2022-12-22
Add lunarphase module to display the phase of the Moon

doctorfree:
The `lunarphase` module displays the phase of the Moon as ASCII art from [Wttr.in](http://wttr.in). No new dependencies are added as this module uses the same dependencies as `prettyweather`.

## Configuration

```yaml
lunarphase:
  enabled: true
  position:
    top: 3
    left: 5
    height: 2
    width: 1
  language: 'en'
  refreshInterval: 5h
```


doctorfree:
The documentation for this module has been submitted in the PR at https://github.com/wtfutil/wtfdocs/pull/168

doctorfree:
Some references to `github.com/doctorfree` were still in the module. I fixed those with an additional commit. Hopefully it passes the checks now.

doctorfree:
The DeepSource check is failing and the error seems to indicate poor style in declaring the `widgets` variable in `app/widget_maker.go`

This is not part of the `lunarphase` module so I am hesitant to include the change in this PR. However, it seems to be the reason the checks are not successful. I've tested the following change to `app/widget_maker.go` and it seems to work (that is, `wtf` continues to work with this change). Perhaps there is a reason this is coded the way it is. Here's the change:

```
diff --git a/app/widget_maker.go b/app/widget_maker.go
index 7176d3fb..6c64f7ad 100644
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -371,7 +371,7 @@ func MakeWidget(
 
 // MakeWidgets creates and returns a collection of enabled widgets
 func MakeWidgets(tviewApp *tview.Application, pages *tview.Pages, config *config.Config, redrawChan chan bool) []wtf.Wtfable {
-       widgets := []wtf.Wtfable{}
+       var widgets []wtf.Wtfable
 
        moduleNames, _ := config.Map("wtf.mods")
```

senorprogrammer:
Agreed, that issue should not be your concern. I'll take a look at that, and thanks for this PR!

PR #1444, opened 2022-12-27
Add keyboard controls to lunarphase module

doctorfree:
This pull request adds keyboard control for the `lunarphase` module to display the next day/week lunar phase, previous day/week lunar phase, enable/disable widget display, and open a site displaying more info on the currently selected day's lunar phase in a browser.

Corresponding documentation will be submitted in a pull request to wtfdocs

### DIFF
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -249,7 +249,7 @@ func MakeWidget(
 		widget = mercurial.NewWidget(tviewApp, redrawChan, pages, settings)
 	case "lunarphase":
 		settings := lunarphase.NewSettingsFromYAML(moduleName, moduleConfig, config)
-		widget = lunarphase.NewWidget(tviewApp, redrawChan, settings)
+		widget = lunarphase.NewWidget(tviewApp, redrawChan, pages, settings)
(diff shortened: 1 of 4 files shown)


doctorfree:
I am cancelling this pull request while I investigate an issue with this module's interaction with other modules. This is related to the hang I described in Discussions. Not sure if the `lunarphase` module is the culprit but do not want to merge these changes until I debug a bit more.

senorprogrammer:
I don't see anything in here that looks suspicious. In theory each module runs in it's own Go routine and so should not be blocking in general and in your case the HTTP request shouldn't be an issue because of that.

doctorfree:
I've gone through the code line by line several times and do not see anything suspicious either. I am being cautious and did not want to submit these changes until I understand what is going on. Thanks for looking at the code. It's puzzling behavior. I'll followup with anything new in the discussion.

doctorfree:
I am re-opening this pull request after having resolved the hang caused by multiple rapid HTTP requests. The resolution of this issue involved a workaround in the module and does not address the larger underlying issue so this problem may still effect other modules.

PR #1496, opened 2023-02-20
Fix gitlab scrolling

### LINKED ISSUE #1459: gitlab module repo scrolling
## What should it do?
When using the GitLab module I can scroll it even when it consists of 1 repo. Other repo pages show the message "Gitlab project data is unavailable". This is quite inconvenient and I prefer that they should loop over/not scroll when it is reaching the end.

Seanstoppable:
Fixes #1459
We used the wrong config key here, which for some reason equates to infinite scroll


senorprogrammer:
DeepSource issue is unrelated to this fix. Looks good to me.

PR #1526, opened 2023-02-27
WTF Manage Tmuxinator Projects

APiercey:
_[Originally discussed here](https://github.com/wtfutil/wtf/discussions/1503)_

# Manage Tmuxinator Projects

Easily manage and open Tmuxinator projects from WTF dashboard.

Screenshot 2023-02-27 at 22 24 17

## To test:
1. Install tmuxinator. E.g. `brew install tmuxinator`
2. Configure `config.yml` like below
3. Create a project: press `n`, provide name, enter. Exit your editor.
4. Refresh the list: press `r`
5. Edit a project: highlight a project, press `e`
6. Open a project: press `enter`

```yaml
    tmuxinator:
      enabled: true
      position:
        top: 0
        left: 0
        height: 3
        width: 3
```

### DIFF
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -86,6 +86,7 @@ import (
 	"github.com/wtfutil/wtf/modules/weatherservices/prettyweather"
 	"github.com/wtfutil/wtf/modules/weatherservices/weather"
 	"github.com/wtfutil/wtf/modules/zendesk"
+	"github.com/wtfutil/wtf/modules/tmuxinator"
 	"github.com/wtfutil/wtf/wtf"
 )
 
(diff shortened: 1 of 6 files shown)


APiercey:
Found a small bug in my code where a white space was being added as a project. It's now fixed.

Seanstoppable:
I gave this a try and noticed a few things:
1. Initial startup of wtfutil with this enabled is VERY slow for some reason on my machine. After some experimentation, this seems to be in conjunction with any cmdrunner modules
2. Creating a new project in wtfutil doesn't show up in the list initially. wtfutil eventually hangs as I go and try to add things and refresh.
3. Trying to execute a project didn't do anything for me. I see an 'exit status code 1'. Trying to do this a few times unfortunately causes wtfutil to lock up
4. Editing a project also doesn't do anything

I am not sure what the behavior is supposed to be for edit and start.

I suspect at least part of this has to do with OS choice. I am on OSX

APiercey:
Hey @Seanstoppable! Thanks for the review. Sorry to hear it wasn't a great experience for you.

Thanks for the feedback! I'll try to make some improvements 👍

APiercey:
@Seanstoppable may I ask, were you running wtfutil inside or outside tmux, and if outside, was tmux running?

Seanstoppable:
Ah, when running inside tmux, I see the behavior makes more sense.
And it looks like several of the problems might go away when that is the case.
Maybe we need to add a check to see if we are running inside tmux for a better experience?

senorprogrammer:
I'm not a fan of any module that requires the app be run inside a specific environment, that'll definitely cause support issues in the future. As @Seanstoppable says, this should ideally check to see if it is inside tmux before allowing anything to happen, and perhaps display a logical error message if it isn't?

The slow startup is likely due to running the commands to `tmux` in the main thread. I recommend moving those to a go routine so as not to impact performance.

APiercey:
👍 thanks for the awesome feedback!

senorprogrammer:
Closing this as stale. Feel free to re-submit the PR if you think this is incorrect.

Issue #1548, opened 2023-03-24
Color display bug on yfinance module


inetAnt:
After upgrading to a more recent version, I identified a bug where the color of rows overlap:

![image](https://user-images.githubusercontent.com/1765366/227525994-502089e6-4795-49eb-86aa-5580f47ec49f.png)

inetAnt:
Expected output (running v0.41.0)

![image](https://user-images.githubusercontent.com/1765366/227527154-390f44e9-ba9f-44a5-a18a-4bd6a544133c.png)

PR #1549, opened 2023-03-24
yfinance: Reset color at end of each row

inetAnt:
We need to reset the color at the end of the row so that the current row's color does not splash over the next one until we set its own color.

Resolves https://github.com/wtfutil/wtf/issues/1548


Seanstoppable:
Thanks!

PR #1572, opened 2023-04-18
feat(config): allow to set the configuration path from `WTF_CONFIG`

noirbizarre:
Hello 👋🏼 

This PR is just adding the possibility to set the configuration file path from the `WTF_CONFIG` environment variable (allowing use cases like by directory dashboard using `direnv`)

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -161,11 +161,18 @@ func (flags *Flags) Parse() {
 		return
 	}
 
-	// If no config file is explicitly passed in as a param then set the flag to the default config file
+	// If no config file is explicitly passed in as a param
+	// then try the `WTF_CONFIG` environment variable
+	// then fallback to the default config file to define the default flag value
 	configDir, err := cfg.WtfConfigDir()
 	if err != nil {
 		fmt.Printf("Error: %v\n", err)
 		os.Exit(1)
 	}
-	flags.Config = filepath.Join(configDir, "config.yml")
+	envCfg := os.Getenv("WTF_CONFIG")
+	if envCfg == "" {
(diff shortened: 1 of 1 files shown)


senorprogrammer:
I like this, good idea.

noirbizarre:
Hi 👋🏼 

Is there something preventing to merge this one ?
As you seem to like this one, if there is any change or improvement needed, I would gladly make them ASAP so you can merge it

senorprogrammer:
Nope, nothing needs to be changed, life just got in the way. Apologies for the long delay.

flags/flags.go
```diff
@@ -161,11 +161,18 @@ func (flags *Flags) Parse() {
...
 		fmt.Printf("Error: %v\n", err)
 		os.Exit(1)
 	}
-	flags.Config = filepath.Join(configDir, "config.yml")
+	envCfg := os.Getenv("WTF_CONFIG")
+	if envCfg == "" {
+		flags.Config = envCfg
+	} else {
+		flags.Config = filepath.Join(configDir, "config.yml")
+	}
```
senorprogrammer, line 177: This looks strange to me - if the os ENV var is empty, it sets `flags.Config` to it, which would make that empty. If it's not empty then it uses the file path?

noirbizarre: My bad, it swapped the statements. Fixed

noirbizarre:
> Nope, nothing needs to be changed, life just got in the way. Apologies for the long delay.

No worries, kind of the same for me.

Anyway, I fixed the swapped statements, it should be OK

senorprogrammer:
Thanks for making this improvement, this is excellent.

PR #1581, opened 2023-05-06
Improve digitalclock module

truman369:
This PR will add some customization options for `digitalclock` module:

- Bold font (like tty-clock util)
- Enable date, UTC and epoch lines separately from each other
- Display date in widget title
- Widget center alignment
widget screenshot


senorprogrammer:
Sweet, thanks!

Issue #1478, opened 2023-02-01
CPU usage increases by 100% on every save of config file


TheJavaGuy:
OS: macOS 12.6 (21G115)
wtfutil version: a63329214c888cfbfc67c7ddcf31887c3c8a1c36 (2022-10-05T15:08:35Z)

I installed wtfutil via homebrew. I started it with custom config (e.g. `wtfutil --config=/Users/ivanmilosavljevic/.config/wtf/ivan_config.yml`). This custom config was just a copy of provided default config. Then I started making changes, like changing positions, width and height of modules, or adding a new module, removing a module etc. After every save of config file the CPU usage of wtfutil went up by 100% (so after 7 saves => 700%).

jkrysakowski:
same here

snekmuffin:
can confirm, same happens using Arch linux and the AUR package.

Vlad-Vargan:
Same for ubuntu 22 amd64 insane CPU usage

jonhadfield:
When wtf detects a change, it restarts the app, reloads the config file, and then starts watching the same file again.  The problem is that the code used to monitor the config file for changes is not stopping the watching processes before the new ones start. This is evident when running `strace` at the time of file change; the number of PIDs running `newfstatat` (system call to get the status of a file) simply double each time.  
To recreate, run wtf and then get the pid:
`ps -ef | grep wt[f]util | awk '{print $2}'`
now watch the rate of calls being made to check for these `newfstatat` calls:
`strace -f -p 1766386 2>&1 | grep .config`
at the same time, modify and then save the config file.
The calls look something like this:
``[pid 1639797] newfstatat(AT_FDCWD, "/home/hadfielj/.config/wtf/config.yml",  <unfinished ...>``

The library used for the file watching (github.com/radovskyb/watcher) seems to be at fault and no longer maintained (no updates since 2019). I'm working on a fix that includes switching it out for fsnotify. Will try to get it out this week for testing.

huyz:
Wow, my M1 macbook burned my lap. First time this has happened

macino:
This is still happening. I've tested it today with version 0.43.0. One CPU core is maxed out and the fans are running wild like they never did before.

opobla:
Hi!

Release 0.43.0 is from Feb 24, 2023, but this issue was fixed on Jan 18, 2024.

[...]

PR #1617, opened 2023-10-26
abort handleRedraws goroutine when reading from closed channel

Crash129:
Fixed https://github.com/wtfutil/wtf/issues/1478

The goroutine is expected to be stopped by closing the channel (see wtfApp.Stop()). But as the channel receive is not checked the handleRedraws goroutine turns into an indefinite loop.


senorprogrammer:
Thank you!

Issue #1615, opened 2023-10-04
PiHole module does not use auth token to get status


Muscaw:
Hey, first of all, I'd like to say I'm super grateful for this repo to exist. I've been setting it up this morning and it's just awesome.

While setting up the pihole module, I noticed that the status does not show up. Basically, there is an unmarshal error when decoding the json.

Upon more investigation, I noticed that the requests is returning a HTTP 200, but the body is an empty json array.

The other requests (top clients and top domains) made by wtfutil to PiHole are working as expected and return expected values.

I'm happy to provide a PR with the fix.

## What's broken?
The pihole module does not send the auth token when fetching the pihole server status. This results in a HTTP 200 response with empty array body. The pihole module tries to unmarshal it, but is ultimately unsuccessful

nebbson:
Cool

senorprogrammer:
Apologies for the delay in responding, and thanks for the PR! I'll take a look and hopefully get it in.

FelicianoTech:
This has been fixed in the `master` branch for awhile now, just not released. I'll close this issue when the next release comes out.

PR #1618, opened 2023-10-26
pihole module: send auth token on summary request

pavhl:
Fixes https://github.com/wtfutil/wtf/issues/1615

### DIFF
--- a/modules/pihole/client.go
+++ b/modules/pihole/client.go
@@ -30,12 +30,12 @@ type Status struct {
 	} `json:"gravity_last_updated"`
 }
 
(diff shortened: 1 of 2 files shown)


FelicianoTech:
Hi @pavhl,

I'm managing this project now. I'm going through all PRs and issues. It looks like this problem was addressed in the PR #1614. I am going to close this PR but I want to thank you for making the attempt.

Pi Hole has released v6, which changes how the API works. WTF only supports versions before v6. If you're still using WTF, would could use help in the near future getting v6 support working. In the meantime, I'll make an issue to update the documentation.

PR #1697, opened 2024-05-14
Healthchecks module: Support title setting, add additional statuses

harpchad:
Two additions to the Healthchecks.io module
- Honor the title setting from common settings.
- Support two additional healthcheck statuses defined in the healthchecks API (New and Paused).


senorprogrammer:
Nice!

PR #1700, opened 2024-07-19
update firewall module to use sudo ufw status properly

docsmooth:
This re-orders the "ufw status" logic to assume it's being run as a regular user, and use "sudo /usr/sbin/ufw status". If the username is root, then instead use simply "ufw status".

This update does NOT take variable elevation commands like pbrun/dzdo/pbsudo into account. I may add that in a future update.

### DIFF
--- a/modules/security/firewall.go
+++ b/modules/security/firewall.go
@@ -44,23 +44,22 @@ func FirewallStealthState() string {
 
 func firewallStateLinux() string { // might be very Ubuntu specific
 	user, _ := user.Current()
+    cmd := exec.Command("sudo", "/usr/sbin/ufw", "status")
 
 	if strings.Contains(user.Username, "root") {
-		cmd := exec.Command("ufw", "status")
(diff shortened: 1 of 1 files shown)


docsmooth:
Why was this marked wontfix?  The UFW status comes up empty on Ubuntu as a regular user without, do I need to reference an issue?

FelicianoTech:
I tried testing out this PR and it hangs my WTF instance. Not sure what's going on right now but I can try to revisit sometime soon.

docsmooth:
@FelicianoTech do you have a NOPASSWD entry in your sudoers for the ufw command? If not, it's hanging for your password

FelicianoTech:
@docsmooth I did not. I would love to figure out a way to notify the user that that's the problem rather than just locking up. I don't want to hold up this PR though so that can happen latter.

Verified this works. Thank you for contributing.

docsmooth:
@FelicianoTech that's a great point. I'll think about this, I have some thoughts

PR #1798, opened 2025-07-13
closes config and file watcher when stop is called.

jonhadfield:
The watchers created in both watchForConfigChanges() (wtf-app.go) and watchForFileChanges() (modules/textfile/widget.go) aren't closed when the application is told to stop.  
This PR adds the config watcher to the wtfApp instance so it can be closed in the Stop function after the widgets are closed. It does the same thing for the textfile widget so it can close the filewatcher when told to quit.  

Not sure if this fixes the raspberry pi issue (hangs after reload) as I'm unable to find the historic PR, but I suspect the leftover watcher may at least contribute to the issue.

### DIFF
--- a/app/wtf_app.go
+++ b/app/wtf_app.go
@@ -31,6 +31,7 @@ type WtfApp struct {
 	pages          *tview.Pages
 	validator      *ModuleValidator
 	widgets        []wtf.Wtfable
+	configWatcher  *watcher.Watcher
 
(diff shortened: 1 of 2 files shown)


FelicianoTech:
@jonhadfield Question, did you request the Copilot review or did this PR just do that by itself?

jonhadfield:
It did it by itself. I do subscribe to Copilot, so maybe it triggers on all PRs automatically.

PR #1799, opened 2025-07-13
Add azlogs module

jonhadfield:
This new module enables table output of logs as queried by a log analytics workspace Kusto query.  
Documentation updates with screenshot will follow in the next few days.


FelicianoTech:
Hi,

Some thoughts here.

1. Can you dive into a bit of detail on how many users would use this module? I'm not familiar with Azure logs. It seems like it parses the results of a file and displays it. Would that data somehow change over time?
2. This PR needs to be rebased and brought down to a single commit.
3. This PR has a conflict.

We should probably walk through the first point before you spend time on 2 and 3.

jonhadfield:
> Can you dive into a bit of detail on how many users would use this module? I'm not familiar with Azure logs. It seems like it parses the results of a file and displays it. Would that data somehow change over time?

In Azure, every service will natively send logs to a service called Azure Log Analytics Workspace. It's the equivalent of an ELK (Elasticsearch, Logstash, Kibana) stack where you, respectively, store, ship, and view/search the logs.  Any users of Azure would benefit from this module by being able to see up-to-date results from predefined queries (Kusto is the name of the query language used in Log Analytics Workspace).  If you're familiar with AWS, this is the equivalent of AWS CloudWatch.  The rate of change is down to the service logging, but it's expected to change every few seconds.

How useful this would be to your user base is entirely down to how many would be using Azure for their cloud computing. If that's small/unlikely, then no need to include this and I'll happily continue with it in my private fork.   

Will rebase asap.

FelicianoTech:
I need to give this a proper review still but I'm aiming to include this in the next minor release.

FelicianoTech:
Just waiting on this to be rebased into a single commit. That would also resolve the current conflict.

jonhadfield:
Now rebased.

FelicianoTech:
LGTM

PR #1832, opened 2025-08-19
Suppress pty errors

darrencauthon:
This PR introduces a "ptySuppressErrors" flag that can be used to hide some of the common pty error messages that can occur.

Common, at least to me.  😄  I think many of these issues are due to the tools, not wtf, but it can be difficult to smooth out the output.  Yesterday I really wanted a graph on my board, and I found this as the only way to make it happen.

image

image


FelicianoTech:
Thank you. This PR will end up in the upcoming v0.48.0 release.

PR #1852, opened 2025-09-11
fix: delete the replace directive by upstreaming the change

### LINKED ISSUE #1698: Allow installation via `go install`
## What should it do?

The current instructions to install via `go get` are deprecated in the recent versions of `go`.

Following the steps in README.md, 
`go go get -u github.com/wtfutil/wtf`gives this error:

```sh
go: go.mod file not found in current directory or any parent directory.
        'go get' is no longer supported outside a module.
        To build and install a command, use 'go install' with a version,
        like 'go install example.com/cmd@latest'
...
```

Running `go install` (which is a great way to install command line utilities) gives this error:

```sh
 $go install github.com/wtfutil/wtf@latest
go: downloading github.com/wtfutil/wtf v0.43.0
go: github.com/wtfutil/wtf@latest (in github.com/wtfutil/wtf@v0.43.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

It would be terrific to be able to install `wtfutil` using `go install`. I have found it to be a reliable way to install command line utilities, in particular.

toby-bro:
Once https://github.com/wtfutil/todoist/pull/5 is closed then this PR should close #1698


FelicianoTech:
I love that you also updated the readme. Thank you.

PR #1854, opened 2025-09-11
feat: diskusage #1713

### LINKED ISSUE #1713: FR: Disk Usage Widget
Like `duf` but simple and neato

toby-bro:
Added a diskusage to close #1713

### DIFF
--- a/app/widget_maker.go
+++ b/app/widget_maker.go
@@ -24,6 +24,7 @@ import (
 	"github.com/wtfutil/wtf/modules/devto"
 	"github.com/wtfutil/wtf/modules/digitalclock"
 	"github.com/wtfutil/wtf/modules/digitalocean"
+	"github.com/wtfutil/wtf/modules/diskusage"
 	"github.com/wtfutil/wtf/modules/docker"
 	"github.com/wtfutil/wtf/modules/feedreader"
 	"github.com/wtfutil/wtf/modules/football"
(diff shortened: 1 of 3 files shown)


FelicianoTech:
Hey, this looks great. Thank you.

I tested locally and it works. I have two related concerns though. The width and height that is needed for this to display correctly.

For the width, I don't think there's much we can do right now. `duf` autosizes its width and it display the smallest it can currently.

For the height, for me after local devices, special devices appear, causing a large height. I don't care to see those always and I would imagine many people wouldn't. I have two possible solutions for this:

1. You can hide the "special" table by default and provide a config setting for people to add it back.
2. Break the widget up into the built-in 'duf' tables. For example, in the config, you would set which table you'd like to display from the duf list of tables: `local,network,fuse,special,loops,binds`. If someone wants to two both local and network for example, they would create two instances of this module in their config and set the table for each.

Thoughts?

Seanstoppable:
Question. Since this just wraps duf, why not just use the cmdrunner module? It is an example of why that module exists.

Its been a while since I was a contributor but for a time I was trying to keep new widgets to be things that actually used interactivity or complex display options. Then would make shell script examples in the cmdrunner docs.

toby-bro:
I agree @Seanstoppable but as there was an open issue I wrote this wrapper... I use cmdrunner for this (and many other) local widgets on my own computer... But you are probably right if we do not want an explosion of command wrappers then this should not be merged...

toby-bro:
@FelicianoTech I agree an optionnal boolean for each of the hide could be added, tell me if you want me to add it (if this PR is to be merged)

FelicianoTech:
1. I think think we should hide all but local at first. We can have a boolean to show others. Speaking of interaction (Sean), perhaps we could only show one table and allow the user to interact with the widget to cycle through other tables (if any). That could be a future PR though.
2. I agree with Sean's sentiment. This is something I have been thinking about. I think there's a lot of modules right now. Some can be merged, some aren't needed. WTF has been going awhile now like this. I will write up a longer post in the future but right now, I am thinking this will be okay up until the v1 release. I hope do have that release within the next few months. Then rules get a bit tighter on what we include and as we build towards a v2 release, what we remove.
3. @toby-bro I'm okay with merging this if maybe in the future there's a PR or two with the thought of adding some interactivity. If you rather continue with cmdrunner and not put much more effort into this module, that's okay too. Just let me know what you'd like to do.

toby-bro:
I think it is not worth merging, we should rather make a library of cmdrunner templates for people who want this sort of widgets

PR #1865, opened 2025-10-09
Update dns.go for additional Linux DNS entries.

alexfornuto:
I am by no means a Go developer, so I'll be up front: This is AI generated code. I asked copilot why I'm only getting two out of three DNS entries returned by my first network device in `nmcli` and it thinks this is the solution, a "fragile split".

### DIFF
--- a/modules/security/dns.go
+++ b/modules/security/dns.go
@@ -36,8 +36,10 @@ func dnsLinux() []string {
 
 	for _, l := range lines {
 		if strings.HasPrefix(l, "IP4.DNS") {
-			parts := strings.Split(l, ":")
-			dns = append(dns, strings.TrimSpace(parts[1]))
+			parts := strings.SplitN(l, ":", 2)
+			if len(parts) == 2 {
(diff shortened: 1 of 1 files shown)


FelicianoTech:
Hey Alex,

I can reproduce the problem. I'm checking through your solution now.

FelicianoTech:
This PR does not fix the problem. After some testing, it turns out the problem isn't in this section of the code at all. When it comes to displaying the DNS entries, the display code manually limits itself to 2 IP addresses. I will create a separate PR in order to fix this issue.

It might make the widget fairly tall for some users, but I will have it show every DNS server configured. Most people will only have 1 or 2 anyway. In the future, I want to introduce some smarts for widgets on when to limit their display and when to output more information.

FelicianoTech:
PR #1866 has been merged to resolve this issue. The fix will be available in v0.47.0 [once it's ready](https://github.com/wtfutil/wtf/milestone/4).

PR #1950, opened 2025-12-15
feat: compact calendar view

voidcontext:
Just wondering if you'd be interested in a compact mode for google calendar:

unknown_001

### DIFF
--- a/modules/gcal/display.go
+++ b/modules/gcal/display.go
@@ -38,7 +38,7 @@ func (widget *Widget) content() (string, string, bool) {
 		}
 
 		ts := calEvent.Timestamp(widget.settings.hourFormat, widget.settings.showEndTime)
-		timestamp := fmt.Sprintf("[%s]%s", widget.eventTimeColor(), ts)
+		timestamp := fmt.Sprintf("[%s]%s ", widget.eventTimeColor(), ts)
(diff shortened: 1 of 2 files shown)


FelicianoTech:
Possibly. Can you please provide more details? Also, I think there's an open issue that suggest the Google Calendar module doesn't currently. I haven't had the time to verify that information though.

voidcontext:
> Possibly. Can you please provide more details? Also, I think there's an open issue that suggest the Google Calendar module doesn't currently. I haven't had the time to verify that information though.

The module seems to be working with v0.46.1, and from the main branch as well.

The idea behind the changes added in this PR is to be able to use the calendar module horizontally rather than vertically, this way each event is displayed in a single line (instead of 2 lines of data + empty line). Maybe the date could be moved into the line as well, like a prefix.

FelicianoTech:
Hey @voidcontext, are you still interested in this?

If so, I like it, but I'm thinking we rework the terminology. Instead of "compact" mode, perhaps there's a vertical and horizontal mode or direction. The default would be vertical for backwards compatibility. Another name can be row vs column mode.

Thoughts?

FelicianoTech:
Closing due to lack of response.

PR #1973, opened 2026-02-06
feat: Add firewalld support

Ferenc-:
[Firewalld](https://firewalld.org/) is a popular firewall management frontend on linux.

This PR adds support for that, which also means that vanilla Fedora/RedHat/Centos/SUSE users will now have out of the box support for `firewall` in the `security_info` module.

### DIFF
--- a/modules/security/firewall.go
+++ b/modules/security/firewall.go
@@ -43,14 +43,14 @@ func FirewallStealthState() string {
 /* -------------------- Unexported Functions -------------------- */
 
 func firewallStateLinux() string {
-	// Check FirewallD first
-	if hasFirewalld := checkFirewalld(); hasFirewalld != "" {
-		return hasFirewalld
-	}
-
-	// Check UFW
+	// Check UFW first
 	if hasUfw := checkUfw(); hasUfw != "" {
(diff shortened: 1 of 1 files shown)


FelicianoTech:
Thank you for submitting this PR. I will get it tested within the next few days. In the meantime, can you address the comment I made?

modules/security/firewall.go
```diff
@@ -41,7 +43,12 @@ func FirewallStealthState() string {
 /* -------------------- Unexported Functions -------------------- */
 
 func firewallStateLinux() string {
-	// Check UFW first
+	// Check FirewallD first
+	if hasFirewalld := checkFirewalld(); hasFirewalld != "" {
+		return hasFirewalld
+	}
```
FelicianoTech: Please move this if block below the UFW block. I'd like to keep the more likely matches towards the top.

Ferenc-: Done

FelicianoTech:
Thanks. After I test this, I'll make sure to get this merged before the next release. That should be before this month is over.

FelicianoTech (approved):
LGTM

PR #1993, opened 2026-02-11
Fix module help: Always include common attributes

MaxGyver83:
Currently, the module help skips the common configuration attributes when module-specific attributes are printed.

Try:

```sh
wtf -m=git
```

Common attributes like `bordered` and `title` are missing.

The reason is that `HelpFromInterface` checks for field name `"common"` but it's actually `"Common"`.


FelicianoTech (approved):
Great catch.

PR #1994, opened 2026-02-11
Make config and module flags non-optional

### LINKED ISSUE #1992: `wtf -m todo` fails
## What's broken?

```sh
$ wtf -m todo
Command `todo` is not supported, try `wtf --help`
```

`wtf -m=todo` works though. This is strange because the used library `go-flags` seems to support arguments like `-m todo`, as well. In their README is an example using `-n Me`.

MaxGyver83:
- This allows `wtf -m <module>` (without `=`).
- Show warning for `wtf -m` (missing module) and `wtf -c` (missing config).

Fixes #1992.

### DIFF
--- a/flags/flags.go
+++ b/flags/flags.go
@@ -16,8 +16,8 @@ import (
 
 // Flags is the container for command line flag data
 type Flags struct {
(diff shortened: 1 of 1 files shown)


FelicianoTech (approved):
LGTM

PR #2028, opened 2026-03-06
Update keyboard_widget.go to process event keys before regular keys

### LINKED ISSUE #2027: Keyboard_widget InputCapture processing order issue when same key used with modifier
https://github.com/wtfutil/wtf/blame/2a48425a6997eaa9943456010f408b0da37ec6d5/view/keyboard_widget.go#L99

Event key needs to be captured and processed before regular key press, otherwise original key press action is performed. 

Noticed in ToDo module, could not demote, promote, or set item to first because keys "<KBD>j</KBD><KBD>k</KBD>, and <KBD>f</KBD>" have their own actions and <KBD>CTRL</KBD> functions were being ignored

solution: 
process keyMap before charMap

```
func (widget *KeyboardWidget) InputCapture(event *tcell.EventKey) *tcell.EventKey {
	if event == nil {
		return nil
	}

	fn := widget.keyMap[event.Key()]
	if fn != nil {
		fn()
		return nil
	}

	fn = widget.charMap[string(event.Rune())]
	if fn != nil {
		fn()
		return nil
	}

	return event
}
```

cbRoy:
Swapped the processing of event presses and regular key presses. This prevents modifier events from being ignored when they use the same defined regular keys, such as<KBD> j</KBD> and <KBD>C-j</KBD>

fixes #2027


FelicianoTech:
Hey @cbRoy. Thank you for the PR.

Did you test this change yourself? If so, on what OS? I'm testing on macOS and it's not fixed for me.

PR #2029, opened 2026-03-08
golangci-lint update fixes

cbRoy:
fixed lint warnings from the updated golangci-lint. removed nil check with len in gcal, and replaced Sprintf with Fprintf in bargraph and uptimekuma modules


FelicianoTech:
Looks good. Thank you. I'm also going to be pinning the version of golantci-lint so that these changes don't happen by surprise anymore.

FelicianoTech (approved):

PR #2051, opened 2026-04-17
Remove CDS modules

FelicianoTech:
The following two PRs have been opened for awhile: #1841 and #1848

Both are related to an issue updated to the new XY dependency, which hasn't been updated with a fix in over a year. I'm not sure how many people are using the CDS modules, or if they currently work, but I don't have much knowledge there and no one has come along to fix these two PRs that have been holding all of this work back.

As a result, I believe the best option is to remove these modules. This will make those two PRs unnecessary, remove a ton of dependencies, and help make the v1 release more supportable.

I will hold this PR open for at least a week to see if there's any objections. I'm aiming to remove a few things before the v1 release so this seems to be in-line.

Closes #1841
Closes #1848


FelicianoTech:
11 days later. Merging.

PR #2069, opened 2026-04-30
Make mempool fee API URL configurable

Bortlesboat:
This keeps the mempool widget defaulting to mempool.space, but lets users set an alternate compatible fee endpoint with `apiURL`.

I work on Satoshi API, so the concrete use case I tested this against is a mempool-compatible hosted fee endpoint:

`https://bitcoinsapi.com/api/v1/compat/mempool/fees/recommended`

The change is intentionally generic: no default changes, no new dependency, and self-hosted mempool.space-compatible URLs work the same way.

Verification:
- `go test ./modules/cryptocurrency/mempool` not run locally because `go`/`gofmt` are not installed in this Windows environment.
- `git diff --check` passed.

### DIFF
--- a/modules/cryptocurrency/mempool/settings.go
+++ b/modules/cryptocurrency/mempool/settings.go
@@ -3,18 +3,20 @@ package mempool
 import (
 	"github.com/olebedev/config"
 	"github.com/wtfutil/wtf/cfg"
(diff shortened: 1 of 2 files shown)


FelicianoTech (approved):
Thank you.

PR #2176, opened 2026-08-06
docker: add pidFilePath to avoid waking a socket-activated daemon

toby-bro:
I noticed that `wtf` would start my Docker daemon again within seconds of stopping it, keeping dockerd running (and burning CPU).

The cause is systemd socket activation (`dockerd -H fd://`) : systemd owns the listening socket and starts the daemon on the first connection, so polling the Docker API to find out whether the daemon is up starts it. 
This PR adds a check that will read the dockerd's pid file, whose default location is documented as `/var/run/docker.pid` instead of querying the socket.
`pkg/pidfile.Read` also verifies the process is alive, so a stale file from a crashed daemon doesn't produce a false positive.

It is opt-in : with `pidFilePath` unset, the module behaves exactly as before. 
If the installation is "standardish" then the `pidFilePath: "auto"` should manage to detect the actual location of the file.
Otherwise it can be manually specified. 

It could reasonably become the default, but a wrong path reads as "daemon not running", and setups like Docker Desktop or a custom --pidfile have no file where we'd look so I didn't want to break anyone's wtf silently.

Unit tests for the new file were written with Claude.

### DIFF
--- a/modules/docker/daemon.go
+++ b/modules/docker/daemon.go
@@ -1,0 +1,44 @@
+package docker
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/docker/docker/pkg/pidfile"
+)
+
+// resolvePidFilePath expands "auto" into a guess at the pid file's location, or
+// "" when there is no local pid file to read. Set the path explicitly if the
+// daemon runs with a --pidfile that isn't next to its socket.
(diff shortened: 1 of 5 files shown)


toby-bro:
I'll open a `wtfdocs` PR as soon as this has "ideological approval"

